PR #7, opened 2019-12-04
Update Rules.md

frablaser:
Section edited:
1- Rule commands
Added:
You can also duplicate the same trigger on many lines. 
`on power2#state=1 do power1 1 endon`
`on power2#state=1 do RuleTimer1 100 endon`

**Rules have a maximum string length**
If the string of your rule is to long and it reach the limit of the rule length, you will loose the end of it... Always verify the status line returned after the insert in the console to see if the last words of your rule has been well inserted. In the case of your rule length is larger then the limit, split it in an other rule<x>...

2- Rule Variables
     Removed the % who seams not needed, see the example in the : 
          You can test it directly in the console like:
          12:44:12 CMD: var4 = utctime
          12:44:12 RSL: stat/tasmota/RESULT = {"Var4":"1575481472.000"}

3- Major features > Multiple Tasmota commands into a ...
   Multiple Tasmota commands into a IF works if statements separated by `;`. For example:  
  `IF (Mem1==0) Var1 Var1+1; Mem1 1;  Delay 10; Power1 ENDIF`
  
   But not like this:
   `Power1 off; LedPower on; IF (Mem1==0) Var1 Var1+1; Mem1 1 ENDIF; Delay 10; Power1 on`
   
   You should split it in two lines like:
   `on power2#state=1 do Power1 off; LedPower on; endon`
   `on power2#state=1 do IF (Mem1==0) Var1 Var1+1; Mem1 1 ENDIF endon`

### DIFF
--- a/Rules.md
+++ b/Rules.md
@@ -125,7 +125,14 @@ Hardware and software serial interface, RF, IR and TuyaMCU are also supported ba
 ### Rule Command
 A rule command can be any command listed in the [Commands list](Commands). The command's `<parameter>` can be replaced with  `%value%` which will use the value of the trigger. 
 
-`ON Switch1#State DO Power %value% ENDON`
+`ON Switch1#State DO Power value ENDON`
+
+You can also duplicate the same trigger on many lines. 
+`on power2#state=1 do power1 1 endon`
+`on power2#state=1 do RuleTimer1 100 endon`
+
+**Rules have a maximum string length**
+If the string of your rule is to long and it reach the limit of the rule length, you will loose the end of it... Always verify the status line returned after the insert in the console to see if the last words of your rule has been well inserted. In the case of your rule length is larger then the limit, split it in an other rule<x>...
 
 To accomplish a rule with one trigger but several commands, you need to use `Backlog`:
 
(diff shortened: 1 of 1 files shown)


ascillato:
Sorry, there are some syntax errors.

meingraham:
I'm reviewing. The issue of some variables not needing to be enclosed by `%` needs to be discussed with the developer (@laurentdong).

ascillato:
In rules code is needed. If % is not in the variable, the rules code won't interpret that.

frablaser:
English isn't my primary language, then, please correct it... ;) 

I tried it and it works without the % ... try it in the console as shown in the example...

meingraham:
https://github.com/arendst/Tasmota/pull/6386#issuecomment-561782820

meingraham:
I think for now I am going to make a couple of your recommended edits but retain the `%` syntax portions until we confirm how variables behave in expressions. I will close this PR w/o merging.

frablaser:
Just for fun...
13:46:55 CMD: var4 = sunset
13:46:55 RSL: stat/tasmota/RESULT = {"Var4":"963.000"}

ascillato:
There are some parts of @frablaser that are Ok to be added to the docs. That is why I was rewiewing it. hehehe. I need some minutes to read and test every example to be added to the docs. I'm not so fast :rofl:

meingraham:
@frablaser yes... when expressions are compiled in. But when not, then the syntax is different. That's the issue. The syntax should be consistent regardless.

frablaser:
I think there is compilation in the rule inserted in the console... the seams to be interpreted.

ascillato:
You are right. Rules are being interpreted, but for using IF you need to compile by yourself adding the optional IF support. None of the precompiled bins include IF support. That is why, the docs must remain as the default provided (for easy understanding for new user) and a separated article or section for the optional IF.

frablaser:
Sure... that's why the IF section become with a warning... 😉

ascillato:
Anyway, give me some time to review it, in order to add your modifications to the docs.

ascillato:
Added your comments with syntax correction plus other added variables.

Thanks @frablaser

blakadder:
I see spelling and capitalisation errors

PR #27, opened 2019-12-22
Update Commands.md

apc300:
command to clear rule


ascillato:
Thanks. :+1:

Changed the 2 `""` for just one. It was programmed for just using one `"`. 2 quotes works but are not necessary.

apc300:
thanks, i was sure you programmed it for something, but i guessed it might be double, which worked better than white space

ascillato:
Theo programmed it ;)

PR #30, opened 2019-12-24
Shutter-and-Steppers.md

TrDA-hab:
Updates: Wiring Diagrams and Bill of Materials.


meingraham:
@TrDA-hab 

Please add a description/use case for each diagram. For example, the parallel setup is to run two shutters from the same controller, etc. How you use them. This will tell users what they are looking at, what options they can consider, etc.

Also, while you're in there editing, please indent the capacitors on the BOM. This was overlooked in earlier edits.

Looks good to merge after that.

Mike

TrDA-hab:
@meingraham 

I will deal with your question. I need some time.

PR #31, opened 2019-12-26
add GPS-NTP-server

Staars:
This looks strange in the preview, that's why I do not want to directly commit it. @blakadder Can you please take a look?


blakadder:
looks ok to me

PR #32, opened 2019-12-27
Shutter-and-Steppers.md

TrDA-hab:
description for diagrams


TrDA-hab:
and i broke again ... indent the capacitors on the BOM !

PR #36, opened 2019-12-29
Update iBeacon-driver.md

lacrimosa2k2:


lacrimosa2k2:
With the Updated Ibeacon Driver File from Gemu2015

PR #39, opened 2019-12-29
Added no neutral 2G version template

IainPHay:
Added details for no neutral version Aliexpress link and added template for 2G no neutral version which is different to that in article.

{"NAME":"Wall Switch 2C","GPIO":[53,52,255,17,18,0,0,0,0,21,22,0,0],"FLAG":0,"BASE":28} works


meingraham:
The template is the same; yes? So the doc should just state that the two devices (neutral & no-neutral) use the same template (as one would expect).

IainPHay:
Should be different I think I’ve reposted the same one. I’ll check shortly, sorry for any confusion

IainPHay:
Yes I’ve made an error in editing and first changed the existing config which should be left as is unless it is confirmed incorrect, I assume not but have no way to test. It was  "Wall Switch 2C","GPIO":[17,255,255,255,0,22,18,0,21,56,0,0,0],"FLAG":0,"BASE":28} and should stay that way.

PR #42, opened 2020-01-02
Initial support for shutter button control

to-scho:
doc for PR #7403


meingraham:
Tasmota supports 8 relays. Why does this command support assigning up to 10 relays?

to-scho:
You are right. Up to 4 shutters are supported, each shutter controls 2 relays.

This command does not assign 10 relays, why do you think so? 

The command can assign up to MAX_KEYS=4 shutter buttons (inherited from support_button.ino). It takes 10 arguments to define how shutters are controlled by corresponding button. Behavior is described in the commit message and there is also a PR in the doc repo.

meingraham:
`Setup button <x> as "up" button for relay <a>`

The write-up made is seem like `<a>` (and thus `<b>..<j>`) were supposed to be relays. If `<a>` is a relay and `<b>` is a different parameter, and `<c>` another, then the documentation should name each parameter accordingly. The command entry can point to the `Blinds & Shutters` section for a full description. But, as it stands, the description in the `Commands` article is misleading.

For this command, does `<x>` correspond to the shutter or to the button? The other commands act on `Shutter<x>`. This command should be structured as "assign button `n` of shutter `x` to perform function `up/down/etc/`. Thus `ShutterButton1 ...` should be assign a button function for shutter `1`. Is that how the command functions?

to-scho:
For ShutterButton <x> is the button number. If you call "ShutterButton3 1 up" you assign button 3 to shutter 1. I will add a list for <x>, <a>...<j> and an example usage as above.

meingraham:
@to-scho 

As all the rest of the command use `<x>` for the shutter, you should reverse the command to follow every other command. Otherwise there will be a great deal of confusion going forward.

@arendst I see you merged the PR. Your thoughts on consistency for the commands?

arendst:
As there can only be four buttons we are running low of them. It would indeed by wise to assign ShutterButton1 to shutter1 etc.

The group functionality then becomes an issue but perhaps to-scho can think of a good solution.

In any case I agree with Mike if we could keep the relation between ShutterButton<x> and shutter <x>.

meingraham:
@to-scho 

In other words, your example would be to use `ShutterButton1 3...` to assign Button `3` to Shutter `1`. This retains the consistency of using `<x>` as the shutter number for all shutter commands.

to-scho:
Current implementation allows to flexibly assign up to four buttons for a single shutter. Inverting will not allow this or will need 40 parameters for the commend. 

The most likely setup is to have two buttons (one for up and one for down) to control a single shutter. A friend of mine has two shutters mounted side by side which are controlled by one sonoff. Due to a lack of GPIOs he only has two buttons to control both shutters. Here you need to assign only a single button (up and down are controlled by multi-press) to each shutter.

Originally I started with the inverted style but ended up in a mess.

meingraham:
From [`Blinds & Shutters`](https://github.com/tasmota/docs/blob/4a5e1ebbcfaaf964984e316c08fb6f28c699ca32/Blinds-and-Shutters.md#button-control)
`ShutterButton3 1` up (same as `ShutterButton3 1 100 50 74 100 0 0 0 1 1`)

This says "assign button `3` to perform the `up` command on shutter `1`".

But instead, to configure the same thing:
`ShutterButton1 3 up` (or `ShutterButton1 3 100 50 74 100 0 0 0 1 1`)

In your case of using a single button to control multiple shutters, can you not repeat the command?
`ShutterButton2 3 up`

How would changing the structure result in needing `40 parameters`? If you can please elaborate with a couple of examples.

Thanks.

Mike

to-scho:
Let's start with 20 ;-) when using two buttons to control a single shutter. With this setup you can use a standard wall dual shutter button. One button is 'up', the other one 'down'. Configuration is done e.g. by "backlog ShutterButton3 1 up; ShutterButton4 1 down". 

For more control about each option this becomes  "backlog ShutterButton3 1 100 50 74 100 0 0 0 1 1; ShutterButton4 1 0 50 24 0 0 0 0 1 1". In total the behavior of the two buttons 3 and 4 is controlled by 20 parameters. You can program up to 8 shutter different positions based on the button press actions. If I now invert the command to ShutterButton1 I need to pass all 20 parameters to that command. I cannot call two times ShutterButton1 to pipe the two sets of parameters. Or I need to have more commands for each shutter. 

Four buttons with 40 parameters for a single shutter is maybe not a real use case.

to-scho:
> In your case of using a single button to control multiple shutters, can you not repeat the command?

I considered having a single button to control multiple shutters is not a use case since this is quite error prone in handling.

meingraham:
I'm not trying to be difficult (it actually comes very easily for me :wink:). I just think that reversing the definition of the enumeration of the command for this single command as an exception to be the button where every other command it is the shutter will be a problem. You wrote the code and you know it. But from a regular user and for those of us supporting Tasmota, I am sure this will be the source of a great number of issues.

```
Backlog ShutterButton3 1 100 50 74 100 0 0 0 1 1; ShutterButton4 1 0 50 24 0 0 0 0 1 1
```

As I stated, yes, you wrote the code so you know the use of each parameter. In your example, I still do not understand why inverting the button/shutter positional parameters causes the parameter list to "blow up". Why does this not work?

```
Backlog ShutterButton1 3 100 50 74 100 0 0 0 1 1; ShutterButton1 4 0 50 24 0 0 0 0 1 1
```
and for a second shutter also using button 3 and button 4:
```
Backlog ShutterButton2 3 100 50 74 100 0 0 0 1 1; ShutterButton2 4 0 50 24 0 0 0 0 1 1
```

The number of parameters for one invocation of the command is still 9 (in addition to the button number).

Even if the button parameter values for each shutter are different, executing the command multiple times allows the same button (i.e., same GPIO) to be used as an input yet the outcome to vary according to which shutter the input is to act on.

Also, some of your parameters use 0/1 to disable where as `<b>..<e>` use `-` to disable. Again, inconsistencies will cause many user issues down the road.

Perhaps the command could take a different approach to parametrizing.

`ShutterButton<x> <button>,<function>,<setting>`
<ul>
<li>&lt;x> &mdash; Shutter 1..4</li>
<li>&lt;button> &mdash; Button 1..4</li>
<li>&lt;function></li>
<ul>
<li>0 &mdash; Disable all buttons for this shutter</li>
<li>1 &mdash; Assign Single press move</li>
<li>2 &mdash; Assign Double press move</li>
<li>3 &mdash; Assign Triple press move</li>
<li>4 &mdash; Assign Long press (hold) move</li>
</ul>
<li>&lt;setting></li>
<ul>
<li>-1 &mdash; Disable button for this &lt;function></li>
<li>up/down/updown &mdash; shutter position assignment for button action, publish MQTT (default)</li>
<li>0..100 &mdash; shutter position assignment for button action, publish MQTT (default)</li>
<li>200..300 &mdash; shutter position assignment for button action (offset by 200), do not publish MQTT</li>
</ul>
</ul>  

```
Backlog ShutterButton1 3 100 50 74 100 0 0 0 1 1; ShutterButton1 4 0 50 24 0 0 0 0 1 1
```
becomes
```
Backlog 
  ShutterButton1 3,1,100; 
  ShutterButton1 3,2,50; 
  ShutterButton1 3,3,74;
  ShutterButton1 3,4,200;
  ShutterButton1 4,1,0;
  ShutterButton1 4,2,50;
  ShutterButton1 4,3,24;
  ShutterButton1 4,4,200
```

This seems like a lot of setup on first inspection... but as this is done only once for the shutter, it's not a huge burden. Also, this command structure allows fine tuning each button's position assignment directly with a simpler command and you don't have to remember the setting for the other buttons to set all nine parameters.

As for the MQTT publishing part
- If a user wants to disable MQTT publishing they cannot use `up/down/updown`. They have to specify the position via the numeric value (`200..300`)
- Alternatively, you can look at [`SetOption66`](https://tasmota.github.io/docs/#/Commands?id=setoption66) or a new `SetOption` to control MQTT rather than using an offset value. Even though the granular ability to control MQTT on an button by button basis seems nice, I think most users will want MQTT either on or off globally.

to-scho:
Thank's for the proposal. I am fine with such serial configuration when we agree about the following two issues. In particular I need your advice for b)

a) assigning a button to more than one shutter will blow up flash configuration space from 16 to 64 bytes. I guess this is out of scope. Anyhow, any assignment can simply overwrite configuration -> the last ShutterButton assignment wins.

b) have Tasmota used such serial configuration anywhere else before? I can't remember that a configuration command do not have an explicit result but depends on the configuration command sequence before. It would not be possible (or code cumbersome) to extract that command sequence from flash. As you stated you not need to remember the setting for all the buttons but you will also not be able to get the configuration stored in flash by calling ShutterButton without any arguments. Configuration commands I used so far always allowed to read the configuration by calling the command without parameters. BTW, this was missing for ShutterCalibration, I found it anoying and committed a change for it today.

meingraham:
[`TuyaMCU`](https://tasmota.github.io/docs/#/Commands?id=tuyamcu) uses a similar approach to configure the different functions of the MCU. You configure one function at a time. But when you execute the command without parameters, it outputs all the configurations.

https://github.com/tasmota/docs/pull/42#issuecomment-570230737
> A friend of mine has two shutters mounted side by side which are controlled by one sonoff. Due to a lack of GPIOs he only has two buttons to control both shutters. Here you need to assign only a single button (up and down are controlled by multi-press) to each shutter.

I probably misunderstood what you meant by this. I thought this was a use case for one set of buttons controlling two shutters. Upon closer inspection, I think what you meant was to leverage the limited number of GPIO by providing an up/down function to a single button. If so, then yes, we can forget about scenario `a`.

stefanbode:
Happy new year together. Highly sophisticated all this requests. This is definitely an expert feature. I have to think if it is still in sync with the different modes of a shutter.

to-scho:
I committed an inverted command suggestion. If you are fine with is I will also update "blinds and shutters" section and finally the code.

to-scho:
> Happy new year together. Highly sophisticated all this requests. This is definitely an expert feature. I have to think if it is still in sync with the different modes of a shutter.

Hi Stefan, ShutterButton adds an abstraction between buttons and the shutter relays. When assigned the shutter will react on the buttons independently from the shutter mode. In particular for shutter mode 1 this is much more women friendly.

In addition you got additional behaviour configuration possibility for double, tripple and hold press activity of a button. MQTT broadcast allows to control all shutters @home e.g. with hold press without any home assistant or what else. 

In particular the hold press feature is women proven by my girl friend ;-)

meingraham:
> I committed an inverted command suggestion.

I cannot find this commit. Please advise.

meingraham:
@to-scho Thank you. I think this will be better in the long run. And differentiating the parameters in the documentation with "descriptive" labels definitely makes it easier for users to comprehend.

If we'd gotten here right away, we would not have explore the "simpler" syntax I suggested. But it's now in my mind... and of course I like it :wink:

@arendst So, now a choice. Two options:
1. Similar to TuyaMCU - https://github.com/tasmota/docs/pull/42#issuecomment-570266665 (would require a re-write - i.e., I drop the grenade in the room and run away :wink:)
2. [As built](https://github.com/tasmota/docs/blob/84c86740666a3a8038091aef6957838a38416212/Commands.md#shutterbutton) by @to-scho

stefanbode:
I already did not like the calibration matrix. Not easy to understand. This feature again has a complex syntax. I’m not sure if we are on the right track. For sure the features are awesome and I also have no better idea.

meingraham:
@stefanbode 

Even though initial button setup would require multiple commands to configure, I think the simpler syntax I suggested is easier for the average user to understand. This only has to be done one time on initial setup. If a user want to make updates, then they have a simpler single command without the "baggage" of the other button setup commands.

`ShutterButton<x> <button>,<function>,<setting>`
<ul>
<li>&lt;x> &mdash; Shutter 1..4</li>
<li>&lt;button> &mdash; Button 1..4</li>
<li>&lt;function></li>
<ul>
<li>0 &mdash; Disable all buttons for this shutter</li>
<li>1 &mdash; Assign Single press move</li>
<li>2 &mdash; Assign Double press move</li>
<li>3 &mdash; Assign Triple press move</li>
<li>4 &mdash; Assign Long press (hold) move</li>
</ul>
<li>&lt;setting></li>
<ul>
<li>-1 &mdash; Disable button for this &lt;function></li>
<li>up/down/updown &mdash; shutter position assignment for button action, publish MQTT (default)</li>
<li>0..100 &mdash; shutter position assignment for button action, publish MQTT (default)</li>
<li>200..300 &mdash; shutter position assignment for button action (offset by 200), do not publish MQTT</li>
</ul>
</ul>  

```
Backlog 
    ShutterButton1 3,1,100
  ; ShutterButton1 3,2,50
  ; ShutterButton1 3,3,74
  ; ShutterButton1 3,4,200
  ; ShutterButton1 4,1,0
  ; ShutterButton1 4,2,50
  ; ShutterButton1 4,3,24
  ; ShutterButton1 4,4,200
```

This is the approach `TuyaMCU` uses. Multiple setup commands yet a single command to provide the summary of all the setup when used without parameters. Also, the current implementation's use of `-` for "reset" is unlike any other Tasmota commands.

You could look at this approach for shutter calibration where each position (30/50/70/90/100) is entered as a separate invocation of the `ShutterCalibration` command.

Mike

to-scho:
oh please no ;-) shutter calibration command is fine like it is. 

The complicated thing is not the command syntax but the real world calibration itself. I have calibrated all my 15 shutters just after XMAS. At the end it took me about approx. 20 minutes per shutter. Most time is taken by doing trial movements using ShutterPosition etc. commands. When you have done all measurements the command ShutterCalibration itself is a no brainer. When you split this in five commands it will just took me more time to type all the commands on my iPad and to repeat commands for all my typos...

to-scho:
To come to a conclusion about the shutter button Command I suggest the committed compromise about the tasmota command table.

The already implemented functionality of the up/down/updown shortcuts should a) be ok for non NERD users and is b) 99% similar to what Mike suggested. 

I will copy the NERD syntax to the shutter and blinds chapter for advanced users when someone is not happy with what I think about an up/down/updown button.

meingraham:
@to-scho 

When one supports users day-in & day-out one longs for the simplest and most concise syntax. Commands with a long string of parameters is usually problematic. Thus converting these two commands to the simpler syntax will help ALL users in the long run.

Mike

to-scho:
Can we close this issue please?
-command for different buttons is inverted/serialized
-disable flag values are harmonized
-basic command is shortened to have only 4 parameters

I don't agree that it's a good idea to split is up further. This will make a simple button setup cumbersome. 

With command ShutterCalibration you give a vector of length measurements in cm. I don't think that there is a better way in giving a vector. Giving each element one by one makes it even more complicated since the command needs to check that vector is in ascending order.

meingraham:
I'm ambivalent toward ShutterCalibration in actuality. Since Stefan brought up he would like to make that simpler, I was just providing a suggestion.

The short ShutterButton 👍 

Mike

PR #44, opened 2020-01-03
Fixed WakeUpDuration

ascillato2:


blakadder:
when did this change?

ascillato:
In the code the limitation was always 3000. We had the value wrong in the wiki. Theo explained that in a recent issue and that limit is a design decision.

PR #49, opened 2020-01-08
Update command.md

GianCann:
A more clear explanation for Interlock command usage.
"Interlock 1" is ambiguous.


meingraham:
`Interlock 1` is the command used to runt on Interlock after the interlock groups have been specified. Your change is not correct.

PR #64, opened 2020-01-25
Update Blinds-and-Shutters.md

lordjaxom:
add example for OpenHAB2 rollershutter MQTT channel configuration

### DIFF
--- a/Blinds-and-Shutters.md
+++ b/Blinds-and-Shutters.md
@@ -285,6 +285,127 @@ cover:
            payload: '{{position}}'
 ```
 
+### OpenHAB2 Support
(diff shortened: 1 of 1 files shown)


meingraham:
I would recommend putting this openHAB information in a separate article and then linking to it from the Blinds & Shutters article as well as the openHAB integration article.

lordjaxom:
Do you have any idea on where to put the separate article?

meingraham:
Same folder, just a new .md file

Jason2866:
@lordjaxom  do you finde some time to add this to the OpenHab section 
https://tasmota.github.io/docs/#/integrations/openHAB

stefanbode:
I see that the transformation map is not required anymore. Instead of sending it to backlog you can send directly to shutterposition1. This now also understand UP/DOWN/STOP and for sure the 0..100.

lordjaxom:
Thanks for pointing that out. I am hereby closing my pull request.

PR #67, opened 2020-01-27
add clarification about calibration

to-scho:


meingraham:
Isn't the unit of measure for `ShutterCalibration` is irrelevant? If I use inches as my measurements instead of centimeters, and enter those for the calibration, isn't what is important the relative position at each percent position?

to-scho:
You are right, the unit itself (cm, inch, anything about the how far the shutter has moved) is irrelevant. What I wanted to clarify is about this PR: https://github.com/arendst/Tasmota/pull/7604. 
Someone thought that for calibration he should stop the shutter at 30%, 50%, ... 90% of the desired shutter movement and then give the time (or uncalibrated %) to the shutter calibration command. 
This is how you use the shutter sethalfway command which is described before. I guess that this confuses the reader. 
-For sethalfway you drive the shutter to a particular output of the calibration and give the corresponding input (uncalibrated % that was needed to get the shutter to exactly the middle position) to the command. 
-For calibration you measure the outputs for fixed inputs and give that outputs to the command. 

Not sure how to describe this better. I thought talking about the units could help here. For sure, it would be better to also mention inch for non metric people. What do you think?

meingraham:
> `ShutterPosition<x>` and record the value **as a percentage of the total opening** (e.g., 63)


> Please note that `ShutterCalibration<x>` takes position measurements (**not** the time it takes to move). During calibration you position the shutter to an indicated percentage (e.g., `30%`) of opening and measure the shutter position (e.g., `15`). Use the same unit of measure for all your measurements (e.g., centimeters, inches, etc.). After calibration `ShutterPosition<x> 30` will move to `30%` opening which will correspond to the position you provided (`15`).

Then the `cm` specification should not be necessary.

to-scho:
perfect

to-scho:
> > `ShutterPosition<x>` and record the value **as a percentage of the total opening** (e.g., 63)

For sethalfway you record the percentage from the console. Maybe 

ShutterPosition<x> to ask for current shutter position. Record the position percentage value from the console (e.g., 63 from stat/<topic>/RESULT = {"ShutterPosition1":63})

meingraham:
> Use `ShutterPosition<x>` to inquire the shutter's current position and record the value. This value is a **percentage of the total opening** (e.g., `63` = 63% of opening).

to-scho:
if no one got's the idea to measure/calculate "percentage of the total opening" from the current cm/inch ;-) I know, makes not much sense, since this should be 50% as by the step before.

meingraham:
One has to assume a few things from the user first being that they have a modicum of intelligence. Bad assumption sometimes; I know :wink: But they are following the procedure step by step and each step builds on the previous. You have to hope that they can "see" the "build".

Also, at the very beginning of the Calibration area in the article there is a link to this tutorial - https://www.youtube.com/watch?v=Z-grrvnu2bU (yes, I know it's a bit outdated now that it's been ported to Tasmota and slightly changed, but the concept is the same).

Hopefully the write-up and the video covers the bases.

to-scho:
right, I have seen Stefans video before usage and got the idea

PR #71, opened 2020-01-28
Added section on clearing a rule

robin13:
I looked for this in the Rules page for quite a while and did not realise it was documented in the Commands section.


ascillato:
Thanks

ascillato:
I have just edited that only one quote is needed. If you use 2 it works too, but only one is enough for deletion.

PR #81, opened 2020-02-07
Addition of Mirabella Genio Powerboard 

knottyau75:
Hi All,
I hope I've done this Correctly.   
Here's my guide on how to convert a Mirrabella Genio Power Board

Knotty


Jason2866:
Thank you for contributing! 
New devices which does not require code change in Tasmota
are added (since Template version is available) only at the Template site.
We would be thankfull if you would contribute here: [New Device Template](https://templates.blakadder.com/new.html)

PR #104, opened 2020-02-15
Update Sonoff-Dual-R2.md (404)

sfromis:
Fixing two 404

(Line 80 not touched by me)


effelle:
It's the no-newline octicon. A trailing newline is missing but should not be an issue.

PR #150, opened 2020-03-01
Add parameters for Novostella RGBCW Smart LED Flood Light

2ZZ:


Jason2866:
Thx for contributing. New devices are added -since Template function is available-
only at the Template repo. It would be great if you add here https://templates.blakadder.com/

PR #152, opened 2020-03-02
Update Rules.md

GianCann:
Added the clarification to line 82: "(Note: however fired when new value is the same as the previously stored in Var&lt;x\> )"


GianCann:
Thank you ;)

PR #181, opened 2020-03-03
Update Rules.md (fixing duplicate anchors)

sfromis:
Unique anchors instead of same name repeated 7 times in a row


ascillato:
Thanks

PR #197, opened 2020-03-05
Update PIR-Motion-Sensors.md

FloatingBoater:
Added specifications of the HC-SR501 and AM312, and experience using the AM312 with a Sonoff Dual R2.

### DIFF
--- a/peripherals/PIR-Motion-Sensors.md
+++ b/peripherals/PIR-Motion-Sensors.md
@@ -26,14 +26,30 @@ optional:
 
 A more [advanced example](Rule-Cookbook#Auto-off-Motion-Sense-Switch) of rules with PIRs.
 
-## AM312 
+## AM312 Human Sensing Module (PIR)
+```
+Operating voltage: DC 2.7-12V;
+Sensor: ≤100 degree cone angle, 3-5 m; (required depending on the lens)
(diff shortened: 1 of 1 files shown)


blakadder:
too technical for the scope of the article

FloatingBoater:
> too technical for the scope of the article

Shame.
As it stands, the article is misleading, as available AM312 devices don't work with ESP GPIOs.

blakadder:
don't work in what way?

Jason2866:
Strang why does it work for me since 2 years?

FloatingBoater:
Hi,

The details are in the rejected update to the docs, but in summary, either a recent batch of 'clone' AM312 has a too high output decoupling resistor (20kR), or a Sonoff Dual R2 Button1 / GPIO09 has too low a pull-up resistor, as directly connecting them as standard doesn't trigger the pin low. 

A DVM confirms the voltage swing is too low for the ESP to see.

The HC-SR501 triggers the pin, but as it's designed for 5V and not 3V3, it false triggers like crazy.

My docs update was intended to help others not hit this problem, and give a fix. I've not (yet) tested with different like, like a bare board ESP12F or Wemos clone.

TTFN,

James

PR #211, opened 2020-03-12
Update Home-Assistant.md

effelle:
Add support for new buttons and switches. 
@blakadder I need your help to review it, thanks.


blakadder:
Not that familiar with the new HA automation systems but it looks ok

PR #223, opened 2020-03-18
Update Configuration-Procedure-for-New-Devices.md (loose anchor)

sfromis:
Making link work


sfromis:
Good catch. I used bad keywords when searching for a replacement anchor. Fixed, thanks :-)

PR #233, opened 2020-03-30
Update deep sleep description

KDMcMullan:
Clarifying use case and language, based on discord discussion with barbudor.


barbudor:
Hi @KDMcMullan 

Thanks for your contribution.
I do agree with most of the changes and I thank you for fixing my frenglish.
I have the following comments :
I noticed 2 typos : 
- missing space between `the` and `deep` in `perform its function during thedeep sleep cycle`
- I believr only 1 `p`is needed in `for the purpose of tempporarily disabling deep sleep`

However I do not agree with the change :
`- ## ESP8266 Deep Sleep Side Effects`
`+ ## WEMOS D1 Deep Sleep Side-effects`
Wemos D1 is nothing more than a ESP8266 with bare minimal support that should be present on **any board**. This includes pull-ups on GPIO0+2 and pull-down on GPIO15. All other pins are not connected so the behavior is intrinsically the behavior of a bare ESP8266.
I suggest to stick with the original title.

As per our Discord discussion, I would suggest to reference Wemos D1 when a reset button is involved. You may replace the original diagram with the attached one and add the appropriate text.

Best regards

Barbudor

deepsleep_minimal

KDMcMullan:
I don't know what went on there. I honestly don't think I changed the title, and I completely agree that D1 is far too specific. I honestly don't recall making that change, nor why I would have done so!!

What does it mean "All checks have failed"?

NB, There's little wrong with your frenglish. Thanks for suggesting fixes to my fat finger typos!

PR #246, opened 2020-04-05
Reflect comit #8056

pcdiem:


arendst:
No need to commit into master. It will be updated on new release only.

In fact the documentation should be done on the tasmota/docs site anyway (https://tasmota.github.io/docs/#/Home)

pcdiem:
Sorry for my ignorance. How do I update the doc? I can't directly edit https://github.com/tasmota/docs/blob/master/docs/PWM-dimmer-switch.md

pcdiem:
For my future knowledge, what would have been the correct way to update 
the documentation?

blakadder:
You did it right its just that i did some edits while you were doing edits. Since the new docs site is operational you can remove the PWM-dimmer-switch.md from the arendst repo

pcdiem:
Thank you. Do you mean PWM_DImmer.md in the root of the arendst repo? 
Should I also remove Device_Groups.md?

blakadder:
Yes you can remove both, just make sure content matches

PR #249, opened 2020-04-07
Update Home-Assistant.md: Added Wifi parameters to component

jtilles:
Added json parameters to the switch component as a demo to show how to get wifi telemetry into component parameters.

Tested configuration and works in my instance
wifi_parameters_tasmota_HA


blakadder:
please split that off as a bonus option, not a default configuration since it can be used on almost any device type

jtilles:
Updated.  The discrete Wifi Sensor was commented out, I added it back and added the json parameters as a optional subset.

PR #255, opened 2020-04-10
Update rules

PablexXXXX:
Just added another metod for two way switch without mqtt

### DIFF
--- a/docs/Rules.md
+++ b/docs/Rules.md
@@ -1795,6 +1795,30 @@ Rule1
 Rule1 1
 ```
 
+Another way to achieve the same without using variables is linking the relays states of two boards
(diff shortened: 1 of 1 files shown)


ascillato:
Hi,

Your example is good but can be simplified to just one rule:

```lua
Rule1 ON Power1#state do websend [192.168.10.181] power1 %value% endon
Rule1 1
```

what do you think?

blakadder:
is this getting edited to a better version or closing?

PablexXXXX:
Hi, thanks for the replays. I will definitely use the one rule using the %value% variable.
I´m not familiar with github. This is the first time I write something. Should I close the pull request?

PR #287, opened 2020-05-06
Add note for not using GPIO 15 with NodeMCU.

edwinm:


blakadder:
This article is not meant for wiring instructions and safe GPIO's are already covered in the appropriate article. 
If you wish you can add the note to the PIR documentation

Issue #292, opened 2020-05-07
Please add the new command to documentation


blakadder:
@ctc
re PR: arendst/tasmota#8243

Please edit [Commands.md](https://github.com/tasmota/docs/blob/master/docs/Commands.md) and include new additions.

ctc:
See: https://github.com/tasmota/docs/pull/299

ctc:
see: https://github.com/tasmota/docs/pull/300

PR #300, opened 2020-05-10
Venetian Blind Support

ctc:
Added description for venetian blind support.
https://github.com/tasmota/docs/issues/292


blakadder:
excellent, thank you

PR #303, opened 2020-05-12
Link to instruction how to build for 512K device

b0rder:
There is instruction on how to successfully build and run Tasmota on 512K device (with WEB interface and minimal modules).


blakadder:
512k build are not officially supported nor we endorse building one or offer any kind of support for it. Whoever is doing that will not come from this page with basic instructions on compiling

PR #306, opened 2020-05-14
Add CC2531 to zigbee.md

kugelkopf123:
Second try.

Please check if the linking of the pictures, especially to the one in the _media folder, are correct. Thanks


kugelkopf123:
Thanks! =)

PR #307, opened 2020-05-14
Reintroduce migration releases table

ThomDietrich:
Hey all,

Not long ago I added a table with static links to all relevant migration path releases to the upgrading article. In commit https://github.com/tasmota/docs/commit/df0f9e87bd39ed338a02f87171eb1ff46495dced which is not linked to it the table was hidden away. Reason unclear. The table includes a few otherwise missing details.
Without the direct links the process of safely upgrading is quite cumbersome and potentially dangerous. If not in a table, the links should be included in some form and I'm open to adapt the PR if we can agree on a better format.

All the best!

### DIFF
--- a/docs/Upgrading.md
+++ b/docs/Upgrading.md
@@ -140,7 +140,7 @@ Remember that you **must take each individual step** between the device firmware
 
 As a safeguard perform "Backup Configuration" before installing a new version. If settings are lost "Restore Configuration" should bring them back.
 
-<!-- | Project Name | Release | Direct Download |
+| Project Name | Release | Direct Download |
 |-|-|-|
 | Sonoff-Tasmota | [v3.9.22](https://github.com/arendst/Tasmota/releases/tag/v3.9.22) | [`firmware.bin`](https://github.com/arendst/Tasmota/releases/download/v3.9.22/firmware.bin) |
 | Sonoff-Tasmota | [v4.2.0](https://github.com/arendst/Tasmota/releases/tag/v4.2.0) | [`firmware.bin`](https://github.com/arendst/Tasmota/releases/download/v4.2.0/firmware.bin) |
@@ -149,7 +149,7 @@ As a safeguard perform "Backup Configuration" before installing a new version. I
 | Tasmota        | [v7.2.0](https://github.com/arendst/Tasmota/releases/tag/v7.2.0) | [`tasmota-minimal.bin`](https://github.com/arendst/Tasmota/releases/download/v7.2.0/tasmota-minimal.bin) |
 | Tasmota        | [latest](https://github.com/arendst/Tasmota/releases/latest) | (Check "Assets" section) |
 
-Follow the path till you reach the latest Tasmota version. -->
+Follow the path till you reach the latest Tasmota version.
 
 Migrating from one minor version to the next is mostly painless as the settings are saved in the same location in flash and newer settings are appended.
 


blakadder:
1. table is missing a few steps which caused issues for users and was therefore removed
2. linking to minimal.bin is not enough since its required to do a full upgrade
3. repeats information already existing

meingraham:
The table Thom created includes links to the Release binary and to the minimal binary.

blakadder:
The way table is designed (and was enforced by the sentence at the end of the table) users downloaded 7.2 tasmota-minimal.bin flashed it, went a version up and soft bricked since after 7.2 there was a configuration change.

ThomDietrich:
> table is missing a few steps which caused issues for users and was therefore removed

Not sure which users you are referring to. Everyone is able to add these missing steps. If something is missing, let's add it but please do not simply delete valuable content without improving on it.

> linking to minimal.bin is not enough since its required to do a full upgrade

It's not quite clear which steps are needed exactly. I didn't run exhaustive tests but following the migration path with minimal binary versions only seemed to work at least for parts of the process. If the recommendation is to upgrade to minimal, then standard, then minimal, then standard.... we should add another column for the standard binaries and make this recommendation clear in the surrounding text.

> repeats information already existing

I would say "summarizes" as the existing article did get a bit out of hand and does not mention the exact version numbers. The most important part of the table are the links to the exact releases, ideally to the binaries.
I can invest some minutes and merge the table and the text below into one, either as a table or enumerated text. @meingraham wdyt?

> users downloaded 7.2 tasmota-minimal.bin flashed it, went a version up and soft bricked since after 7.2 there was a configuration change.

That's not good! Let's add this to the table then.

blakadder:
> > table is missing a few steps which caused issues for users and was therefore removed
> 
> Not sure which users you are referring to. Everyone is able to add these missing steps. If something is missing, let's add it but please do not simply delete valuable content without any improvement.

you're not sure which users i'm referring to? cute...

valuable is a very wrong word for incomplete and misleading
> 
> > linking to minimal.bin is not enough since its required to do a full upgrade
> 
> It's not quite clear which steps are needed exactly. I didn't run exhaustive tests but following the migration path with minimal binary versions only seemed to work at least for parts of the process. If the recommendation is to upgrade to minimal, then standard, then minimal, then standard.... we should add another column for the standard binaries and make this process clear in the surrounding text. Could someone confirm this please?

If its not quite clear to you how do you have the gall to recommend steps that can potentially brick other devices? there are people that do extensive tests and convey their information so I'd rather take their information.

 > > repeats information already existing
> 
> I would say "summarizes" as the existing article did get a bit out of hand and does not mention the exact version numbers. The most important part of the table are the links to the exact releases, ideally to the binaries.
> I can invest some minutes and merge the table and the text below into one, either as a table or enumerated text. @meingraham wdyt?

you can say whatever you want, if you want to summarize do a youtube video.

> 
> > users downloaded 7.2 tasmota-minimal.bin flashed it, went a version up and soft bricked since after 7.2 there was a configuration change.
> 
> I suppose version 8.1 or 8.2 did not yet exist when I first drafted the table. It just needs to be added. Fairly unavoidable...

Not really, all you had to do is inform yourself and complete the table.

ThomDietrich:
Wow. You don't learn do you? How can one be so self-righteous!? Very conscious of the fact that you are difficult to work with I offered to add improvements as we surface them. You shout out insults and close the discussion. This is not the way things work.
From now on please do not involve yourself in tickets where I am involved and let other maintainers handle business professionally.

@arendst this guy is the worst to have any power over any community. I've contributed a lot to the documentation when it was still maintained in the GitHub wiki. I do not have that amount of free time any longer and might not invest the potentially wasted time to add further contributions in the future. Judging by how other conversions went I'm pretty sure there are many others feeling the same way.
I am a big fan of your work and will stay a passionate user of the Tasmota project. Thanks for your amazing work!!

arendst:
I followed this issue with increasing dismay too. I haven't paid much time on the documentation as others seemed to do it much better.

I also know you, Thomas have contributed a lot to the initial documentation which now is disabled in favour of a more searchable solution.

Considering the current matter I agree that having links to the correct released versions makes migrating more user friendly than just providing a link to the Github release page which, by now contains over 70 releases.

Although we all know users hate to read and the current page is almost too long for anyone junger than aged 50 I would suggest to @blakadder to add these links back including both the -minimal AND the available released smallest binary (which will update the settings to it's version number). As a compromise it may even be a link to a step-by-step page to keep the initial view of the page within the adolescent comprehension.

blakadder:
> add only the links to the appropriate release page in the list describing the migration path

i have already suggested that but no changes to the PR have been made

arendst:
Changes will be made to the migration document in due time providing the requested information.

meingraham:
I recently ran in to the issue of upgrading several 6.6.x device's and found it difficult to locate the binaries for the required interim versions. I consider myself a fairly experienced user so I imagine less experienced users would find this very taxing and are likely to make errors.

I no longer contribute to the documentation for similar reasons as noted above. I also find informative detail is removed in lieu of purely asthetics. In my case I opted to record the migration path along with links to the required binaries in my own personal notes.This is not adding any value to the community but it has lowered my blood pressure.

PR #312, opened 2020-05-18
Replace incorrect apostrophe with backtick in docs

ajschmidt8:
This PR replaces a misplaced apostrophe with a backtick in one of the doc pages.


blakadder:
Good catch

PR #357, opened 2020-06-05
Update Peripherals.md

pmknowles:


pmknowles:
I'm not sure whether you are OK with links to other sites but these are useful
https://randomnerdtutorials.com/esp8266-pinout-reference-gpios/
and
https://randomnerdtutorials.com/esp32-pinout-reference-gpios/

PR #368, opened 2020-06-20
TuyaMCU updates for current log format

sillyfrog:
Some updates to the displayed log formats that I had when using the web interface. I found them easier to use, and thought it would be of value to others to update the docs to reflect the changes.


blakadder:
great! thanks a lot for that update

PR #374, opened 2020-06-27
Link to TuyaMCU bookmarklet

sillyfrog:
There didn't appear to be any concerns in the Discord channel with linking to the bookmarklet I did for the TuyaMCU, so linking this in on the relevant page.


blakadder:
Could also use a link in Configuring unknown devices! ;)

PR #384, opened 2020-07-08
Update DS18x20.md that it is only in tasmota-sensors.bin

foxsam21:
Added that only available in tasmota-sensors.bin. Also I am not familiar with the other sensors but maybe there are others that also need updating.

### DIFF
--- a/docs/DS18x20.md
+++ b/docs/DS18x20.md
@@ -1,5 +1,7 @@
 # DS18x20 temperature sensor
 
+!!! info "This feature is included only in tasmota-sensors.bin"
+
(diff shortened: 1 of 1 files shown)


curzon01:
That is not correct:

- USE_DS18x20 is listed in [Available Features and Sensors](https://github.com/arendst/Tasmota/blob/v8.3.1/BUILDS.md) for Tasmota.bin too (Release v8.3.1)
- the official release v8.3.1 tasmota.bin at https://github.com/arendst/Tasmota/releases/download/v8.3.1/tasmota.bin (.gz) still supports DSx1820 (just testet)

don't know where your experience come from but DS18x20 is definitely in

sfromis:
I've also verified that DS18B20 works on the base tasmota.bin, both on 8.3.1 and 8.4.0

PR #385, opened 2020-07-08
RF/IR Repeater rule added

Adityakush:
I was looking for an RF repeater and @s-hadinger helped me to get to this rule to realize that. @ascillato suggested that this use case seems interesting and should be added to the rules cookbook. 
Thanks


blakadder:
thank you, it is indeed an excellent contribution to the rules

Adityakush:
Hi @blakadder. The color of the rule text came out to be all red on the tasmota/docs page. Can you please fix that. Thanks

PR #390, opened 2020-07-09
Fix doorbells add1 to var1

cniweb:


ascillato:
Sorry, the add command is correct. Please, read the explanation that follows those rules.

cniweb:
OK, my mistake, sorry.

PR #407, opened 2020-08-01
Update DS18x20.md to inform about max of 8 sensors

fluppie:
Update DS18x20.md to inform about max of 8 sensors


blakadder:
thanks

PR #410, opened 2020-08-06
Update Buttons-and-Switches.md

haniham:
Added a solution which took me quite some time to figure out.
Application would be a sonof mini with an additional added reed contact, which should transmit instant updates with regular repeated messages, but not interdere with switch1-->relay1.


blakadder:
Please add this way of use to rules cookbook instead

haniham:
Done #412

PR #418, opened 2020-08-20
Update Peripherals.md

tichris0:
Added the mapping between GPIO and ESP8285 pinout to help with hardware mods & set the GPIOs as the first column for readability


meingraham:
Unless the ESP8266 and ESP8285 have the same pin-out, I would recommend not adding the column at all. If they are **_identical_**, then the column heading should not specify `ESP8285`.

PR #425, opened 2020-08-27
update supported commands for BH1750

device111:


blakadder:
Sorry but your code editor made too many changes to verify what's actually changed

PR #440, opened 2020-09-10
Adding Sevo motor use.

TrDA-hab:


TrDA-hab:
@stefanbode Take a look, please. Perhaps some fixes are needed.

elik745i:
Would be good to help the fella!!!

stefanbode:
The servo has no direction. The direction is also controlled  by the Pwm1. The relay 2 shows the direction but it is not used. Therefore we have to be more clear, otherwise people all how to write the direction into the servo. Additionally, we have to add pwmrange during setup. People have to define min and Max because every servo is different here

TrDA-hab:
@stefanbode 

>**In the xdrv27 you will find hardcoded the min and max pwm for the relay.** Please use pwmfrequency 200 for the relay.
you have to set shuttermode 5 to enable the new mode for servos. Additionally, you can define shuttermotordelay e.g. 1.0 for 1 second to have a smooth ram of the velocity. open/close is always full operation 180°. openduration and closeduration can be set to limit the speed in general. Please have fun and let me know if it also works for you.

I beg your pardon, but this is the only clue - how the servo works. The rest I just figured out myself.

_Originally posted by @stefanbode in https://github.com/arendst/Tasmota/issues/8901#issuecomment-687644930_

> Additionally, we have to add pwmrange during setup.

If I understand you correctly, then changing **pwmrange** will give nothing to the user.

The rest is almost clear to me, I must think a little.
I cannot change my PR, and I beg you:
- or reject it (and I will quickly make a new one)
- or accept (and I will make quick additions)
- or I ask you to make the necessary additions.

TrDA-hab:
I have added all the necessary changes.

TrDA-hab:
@stefanbode 
I will be very grateful to you, and if you have a consolidated time, please rate my work for the community. Thank you in advance!

kugelkopf123:
I just uploaded the current dev version of Tasmota on a Shelly2.5. Connected is a standard roller shutter motor. One relay up, one relay down. 
In my house I have used this configuration 7 times. All of them still run on 8.3.1. I would like to think that after 7 pieces and several nights of work, I have understood how to calibrate them. For test purposes I have now installed the current version of Tasmota on one of them. 
I did a reset and tried to calibrate this shutter. I did it like with all other attempts, but then the roller shutter does not move at all.
So I did it again according to the exact instructions of the documentation. And also this does not lead to any result.

It absolutely has to be revised. I don't know if something has simply changed in the calibration procedure or if it is a bug in the shutter driver. 

Following error:

Given is a completely virgin Shelly2.5. system with a correct template:
(SETOPTION) SO80 is off.
Interlock 1
Time measured for up and down. (13s / 12.9s)
Shutter completely lowered.
SO80 1
Restart 1
shutterrelay1 1
shuttermode 1
shutteropenduration 13
shuttercloseduration 12.9
restart 1
Shutteropen1
-> Roller shutter moves up completely.
Shutterclose1
->Roller shutter closes completely.

Great so far. Now Halfway:

Shutter position1 67
->Roller shutter moves to half of the window. (Half height.)
shutterclose1
-> Shutter closes completely.
Shuttersethalfway1 67
restart 1
Shutteropen1
-> roller shutter only moves to an actual height of about 1/3 of the window. So 2/3 of the window is free.


Exactly that, was not in the past. After this command, the 50% position was correct and the roller shutter still opened and closed completely.

In the description it is now said that it is possible to add another second of motor running time. But this cannot be the sense of the whole calibration.

stefanbode:
I will check the calibration. Maybe something wrong here but no change in the code as far as I know.

stefanbode:
Agree! A bug in the calibration calculation. Will be fixed today

TrDA-hab:
@stefanbode 

If you need information for a FREE Tasmota community (no slaves), you can find it here (no, I will not upset with your behavior, I always trust you, and I understand everything - you are a hostage, and I am very sorry, thanks for your work, I'm sorry - they cheated on you again):
https://github.com/TrDA-hab/Projects/blob/master/Servo%2BESP8266/README.md

PR #457, opened 2020-09-20
Update LC-Technology-WiFi-Relay.md

aniva:
Changes for 4x relay to return to the same state after power cycle and some minor fixes

### DIFF
--- a/docs/devices/LC-Technology-WiFi-Relay.md
+++ b/docs/devices/LC-Technology-WiFi-Relay.md
@@ -41,29 +41,40 @@ To configure an LC Technology ESP8266 Relay X2, use the following settings...
   ```
 * Enable the rule (type `rule1 1` in the Tasmota console)  
 
-## LC Technology WiFi Relay - Quad Relay (note, older versions of this board used a baud rate of 9600, so if 115200 doesn't work, try 9600)
+## LC Technology WiFi Relay - 4x or Quad Relay (note, older versions of this board used a baud rate of 9600, so if 115200 doesn't work, try 9600)
 
 Note: The template provided below did not work on an ESP-01 running Tasmota 8.1.0. It was necessary to manually enter the template in the `Configure Template` menu.
 
-* In configuration open `Configure Other` paste this template and select activate
-`{"NAME":"LC Technology 4CH Relay","GPIO":[52,255,17,255,255,255,255,255,21,22,23,24,255],"FLAG":0,"BASE":18}`
 * Open `Configure Module` and set GPIO0, GPIO2, GPIO4 and GPIO5 as Relay1, Relay2, Relay3 and Relay4. Click Save.
 * Disable SerialLog (type ``seriallog 0`` in the Tasmota console)
 
 Enter this command in console (configure the 1st rule)  
 ```
 Rule1
- on System#Boot do Backlog Baudrate 9600; SerialSend5 0 endon
- on Power1#State=1 do SerialSend5 A00101A2 endon
- on Power1#State=0 do SerialSend5 A00100A1 endon
- on Power2#State=1 do SerialSend5 A00201A3 endon
- on Power2#State=0 do SerialSend5 A00200A2 endon
- on Power3#State=1 do SerialSend5 A00301A4 endon
- on Power3#State=0 do SerialSend5 A00300A3 endon
- on Power4#State=1 do SerialSend5 A00401A5 endon
- on Power4#State=0 do SerialSend5 A00400A4 endon
+on Power1#Boot=1 do RuleTimer1 5 endon
+on Power2#Boot=1 do RuleTimer2 6 endon
+on Power3#Boot=1 do RuleTimer3 7 endon
+on Power4#Boot=1 do RuleTimer4 8 endon
+on Power1#State=1 do SerialSend5 A0 01 01 A2 endon
+on Power1#State=0 do SerialSend5 A0 01 0 0A1 endon
+on Power2#State=1 do SerialSend5 A0 02 01 A3 endon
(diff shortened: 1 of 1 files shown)


docs/devices/LC-Technology-WiFi-Relay.md
```diff
@@ -41,30 +41,41 @@ To configure an LC Technology ESP8266 Relay X2, use the following settings...
...
 
+Then enter this command in console (configure the 2nd rule)  
+```
+Rule2
+on Rules#Timer=1 do SerialSend5 A0 01 01 A2 endon
+on Rules#Timer=2 do SerialSend5 A0 02 01 A3 endon
+on Rules#Timer=3 do SerialSend5 A0 03 01 A4 endon
+on Rules#Timer=4 do SerialSend5 A0 04 01 A5 endon
+```
+Enable the rule (type `rule1 2` in the Tasmota console). Note that second rule and timer in the first rule required so that 4x relay can "remember" state after power cycle.
```
aniva, line 77: must in face be `rule2 1`

blakadder:
I don't undestand what you're trying to achieve here. Also you're using GPIO's you shouldn't be using for relays

aniva:
Have you setup 4x relay yourself, do you understand principle of how it is
controlled over serial? Let me know if you want me to explain?

aniva:
Looks like you are more interested in having your way rather one helping community to have less troubles in setting LC switches , lol

meingraham:
@aniva 

Please post a link to the device you are controlling?

Our experience with LC devices in the past has been that control of the relays is done strictly via the byte strings sent to the MCU over the serial interface (i.e., via the `SerialSend5` commands). Your change to the GPIO configuration to add "directly connected" relays, is different and puzzling as to why there are MCU commands AND relay GPIO required.

Your documentation update will break the configuration for the "older" devices for which this article was originally written. If your device is indeed different and requires this different configuration, then any edits to this article should retain the previous configuration and your updates should note that it is specific for your device type.

aniva:
Hi Michael,
Thanks for detailed comment. Kindly highlight where do I change to GPIO
configuration? The only change is to script and comment specifically for 4x
LC relay which with current config (as in the current version of document)
will not work properly.
Regards,
Andrei

meingraham:
The removal of the template...
image

aniva:
Yes, this was in the original text that I removed, this is not my change it was there before.

meingraham:
That's my question, why did you remove the template configuration?

aniva:
Thanks for clarifying. To be sure, I just configured a new module from
scratch. Yes, templates are not needed.
Module communicates with 4x relay over GPIO1 which is Serial Out. At any
moment, and without any config, the relay will be triggered by for example
"SerialSend5 A0 01 01 A2" in the console. Serial command specified by maker
of the relay http://www.chinalctech.com/cpzx/Programmer/427.html   (would
be good to include link, by the way).
Only requirement is to configure other available GPIO, as Relay's (power),
for example
GPIO1 Relay1, GPIO2 Relay 2, GPIO4 Relay3 and GPIO5 as Relay4
In this state, toggle will power corresponding relays. Last step is to
activate a rule that will send over serial to the open/close actual relays.
Now, another problem with config in the current doc is that upon power
cycle the relay will NOT return to a "before state". For this reason, I
added additional rules and tested - works like a charm, upon power back,
relays goes into the before state. Let me know if you have any questions.

aniva:
It should be GPIO0 Relay1 (typo)

Jason2866:
Thats what templates are for to define GPIOs

aniva:
1. Template in the doc is not correct 2. Not reliable - different result
and experience for different users 3. User must understand what is done and
what is the result, and blind copy paste template does not help the
understanding and there may lead to setup that does not work or harder to
troubleshoot

Jason2866:
Template "just" set GPIOs. Not more not less. Has nothing to do with reliability.
Using Template is the standard way of configuring devices. Needed rules are added in the 
description.

aniva:
Configure template “just” set GPIO too, bottom like it more universal ,
when I talk about reliability is an talking from user experience, better
experience better outcome (implies more reliable result) but again fix the
template and address what matters, my comments on script itself, current
version the doc is broken

Jason2866:
Your point of view. Tasmota is designed to work with Templates for configuring.
Many devices needs a special base template to work. You cant achieve needed functions
just with configuring gpios and rules. So the way it works is template and if needed added setoptions and rules. Thats the guideline for contributing to the docs

aniva:
My LC 4x works, any user who uses this doc will struggle, your point of
vote vs my point of view, but leaving it as is does not help user ... when
did you confirm LC relay last time?

meingraham:
@aniva 

I'm not questioning that your configuration steps work for the device you are working with. But you cannot assume that the documentation that another user contributed for their device is wrong. What should happen is to state that for the specific device you are working with (with specific links to it, photos, PCB ink screened model and revision numbers) then your steps apply.

As for the use of a Module vs. a Template - the Template mechanism was adopted almost two years ago to provide a way to support hundreds of devices through extended flexibility. Sometimes using one of the predefined Tasmota Modules will suffice. But anything that can be done with one of the Modules can be done with a Template and them some. It is just a matter of understanding the use of a Template and using the appropriate BASE and GPIO.

Thus, if you are willing to update the documentation to state your specific situation and leave also the previous configuration (perhaps with a note stating that newer devices are now on the market), then that will be considered for merging into the documentation. Otherwise, it won't. Based on the many years of experience represented by the people who have entered in this dialog with you, we respectfully ask you to consider adapting your position. Of course, you are welcome to retain the information that works for you in your own notes.

Regards.

aniva:
Ok perfect, thanks for constructive feedback , I will update changes as
you suggested, cheers

PR #474, opened 2020-10-06
Update Commands.md - CORS default value incorrect OR bug in code

JanKerstens:
Based on a OWASP ZAP scan, the default value for CORS is 
* = enable CORS for all locations

So either the doc is faulty, or there is a bug (i think it is a bug)
" sets an empty field, 
* sets a * field.


blakadder:
What is this PR for? Changing the title?

JanKerstens:
Sorry for the confusion - it is not about changing the title  (i am not yet very qualified on github, more PM expperience on JIRA :-) ).

it is about the WiFi chapter - CORS parameter
in the current version it states            " = disable CORS (Cross Origin Resource Sharing) (default)
but when checking with ZAP, it seems as if the * = enable value is the default.

I am not capable to dive into the code to see if it is a bug or if it is a doc issue.
If you direct me to another page, I can post there if required.

JanKerstens:
should have looked better - closing the pull request and opening an issue

PR #478, opened 2020-10-10
Rework openHAB tutorial

stefan-muc:
I reworked openHAB tutorial to only reflect MQTTv2 binding ([released about two years ago](https://www.openhab.org/blog/2018-12-16-mqtt-arrives-in-the-modern-openhab-2-x-architecture.html)) - one really should start using that, it's out long enough and even openHAB 3 being on its way. Old MQTTv1 tutorial still linked for references, but it's really not recommended to use it (and both bindings can be run in parallel, just in case).

I'm unsure if indenting for code examples is good this way, I thought it might help reading it.

See individual commits for details on what I have done - I suggest not doing a fast-forward merge. I only did (maybe) atomic commits to structure what I did and make a review easier.

### DIFF
--- a/docs/openHAB.md
+++ b/docs/openHAB.md
@@ -48,7 +48,7 @@ Bridge mqtt:broker:myMQTTBroker [ host="IPofBroker", secure=false, username="myU
 Switch Switch_TH "Switch_TH"  { channel="mqtt:topic:myMQTTBroker:tasmota_TH_Thing:PowerSwitch" }
 
 String  Switch_TH_Temperatur "Temperatur [%s °C]" <temperature> {channel="mqtt:topic:myMQTTBroker:tasmota_TH_Thing:Temperature"}
-String  Sonoff_Version "Tasmota Version: [%s]" <tasmota_basic> { channel="mqtt:topic:myMQTTBroker:tasmota_6_Thing:Version"}
+String  Tasmota_Version "Tasmota Version: [%s]" <tasmota_basic> { channel="mqtt:topic:myMQTTBroker:tasmota_TH_Thing:Version"}
 ```
 
 **.rules File for the Maintenance Action:**
@@ -56,9 +56,9 @@ String  Sonoff_Version "Tasmota Version: [%s]" <tasmota_basic> { channel="mqtt:t
 ```js
 // Work with a list of selected Tasmota modules
 val tasmota_device_ids = newArrayList(
-    "tasmota-A00EEA",
+    "Tasmota_A00EEA",
     //… add all your modules here!
-    "tasmota-E8A6E4"
+    "Tasmota_E8A6E4"
 )
 // OR
 // Work with the grouptopic, addressing ALL modules at once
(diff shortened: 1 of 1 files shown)


blakadder:
thanks, looks good!

meingraham:
There are a ton of openHAB user still using the mqtt1 binding. The mqtt2 binding still has some quirks that have not been fully resolved. I would recommend _**not deleting content**_. Perhaps it can be set off to a separate article and referenced in the "main" article via a link.

blakadder:
or simply revert the commit for now?

stefan-muc:
I don't see the problem with MQTTv1 - I provided a link to old documentation, so it's just a click away, nothing deleted.

If one really wants to have it in same layout, one can truly write a separate article.

I'm just not with you about MQTTv1/MQTTv2 differences - only MQTTv2 supports openHAB Thing/Item structure that was introduced with openHAB 2, it has huge advantages and I'm using it since months. I can't see quirks that would stop the majority of users from using it, especially not for Tasmota.

stefan-muc:
>  for users that have not been able to migrate to mqtt2.

And by the way: You don't need to migrate your whole configuration to MQTTv2 to be able to use MQTTv2 - v1 and v2 can be run in parallel.

meingraham:
> v1 and v2 can be run in parallel

Mostly... but there are a couple of issues. This is part of what's kept me on v1.

mqtt2 has pros & cons. Yes, Things, Channels, Items, etc. are more "thorough". If one is setting up OH now, definitely one should start with mqtt2. I started with mqtt1 and have been making my way to mqtt2... but again, there have been issues. Particularly conflicts with some libraries (Jackson, etc.) that are also used by other bindings (e.g., Chromecast). Also the Broker Thing when using the external broker many times goes offline. At least when trying to run both bindings. Also, the mqtt2 developer does not want to extend the binding to support .things and .items text configuration files. I don't care for PaperUI and I try to avoid using it for configuration as much as possible.

Anyhow, if you've moved the Tasmota mqtt1 documentation and example content to a separate link, that suffices for those cases when a user needs it. In any case, OH3 is around the corner so there is likely to be an updated binding... and perhaps the Tasmota mqtt2 documentation will have to be archived as well... in a couple of years :wink:

stefan-muc:
> Also the Broker Thing when using the external broker many times goes offline. At least when trying to run both bindings.
Interesting - I don't experience this behavior, having both versions running since months.

Do you use separate clientIDs for every binding (or let OpenHAB decide)? If they are the same, surely each binding will kick out the other one as the clientID has to be unique. But that's MQTT details a binding can't overcome.

> Also, the mqtt2 developer does not want to extend the binding to support .things and .items text configuration files. 

He doesn't? But what's with the examples in this tutorial and the way I use MQTTv2 binding? I only use *.things and *.items text configuration files to configure MQTTv2 binding (commited to a Git repository). It's wrong information that MQTTv2 doesn't support textual configuration.

>I don't care for PaperUI and I try to avoid using it for configuration as much as possible.

Same for me - as it also will be obsolete for openHAB 3 as far as I understood?

> Anyhow, if you've moved the Tasmota mqtt1 documentation and example content to a separate link, that suffices for those cases when a user needs it. 

Okay great, thanks for your work!

meingraham:
It's been several months since I tried using mqtt2. At the time text configuration files were not supported. I'd been looking at using the REST Docs to bulk load MQTT Things & Items and did some testing. But then I ran into those other issues. Glad to hear someone has added the "standard" text files capability to mqtt2.

PR #482, opened 2020-10-13
Fix incorrect template for wESP32

nagyrobi:
See https://github.com/arendst/Tasmota/commit/37886f33c87d401d0eef98de9a6306ec5764793d


blakadder:
thanks

PR #483, opened 2020-10-14
Update HM-10.md

cayco:
I've added some details on setting up the wiring for HM-10 and configuration of the baud rate.


blakadder:
if you use esp8266 is vague
selecting generic module is optional and depends on the device you're connecting HM10 to

PR #484, opened 2020-10-14
Update PAJ7620.md

VivantSenior:
Just a small clarification


Staars:
I suggest to use „not used“ like for the APDS9960 in order to unify the naming conventions.

PR #523, opened 2020-11-24
Update Rules.md

reimerp:
don't exactly know if this will trigger, but for sure Timer4 is undef. Maybe you need to create a forth one to disable rule1 at sunset. W/o this the Power stays on due to the rule during night even after Timer2

### DIFF
--- a/docs/Rules.md
+++ b/docs/Rules.md
@@ -707,7 +707,7 @@ Backlog switchmode1 1; rule1 1
 ```haskell
 Rule1 ON switch1#state=1 DO Backlog Power1 on; ruletimer1 0 ENDON
 ON switch1#state=0 DO ruletimer1 300 ENDON
-ON rules#timer=1 DO Power1 0 ENDON
+ON rules#timer=1 DO Power1 off ENDON
 ```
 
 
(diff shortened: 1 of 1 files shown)


ascillato:
Hi,

Thanks for your suggestions but they are not correct. Sorry.

The first one is not needed. POWER1 OFF and POWER1 0 are the same thing.

The second change is wrong. The explanation that is above the change is talking about timer1 and not timer4.

reimerp:
Hi, 
1. I know that "Power1 OFF" and "POWER1 0" are the same, but I find it hard to understand and not a good style btw to mis both declarations in one rule. Might confuse especially newbies.

2. Timer4 should disable the rule and its missing (!). I think the intention is to
- ask the sensor during daytime
- switch on at sunset and disable the sensor
- switch off at midnightt

The current solution, does NOT deactivate the sensor, it only activates it by timer. So it stays acrive during night even it it is switched off ad midnicht by Timer. So the off time make no sense at all

meingraham:
I agree, there is no timer #4 in write-up of the setup. It's likely a typo and should be time #1.

Recall that Rule Sets are size limited (less of an issue with compression). This is why `0` is used in rules.

If a user is advanced enough to use rules, one would hope that they have a good understanding of the commands, particularly `POWER`. In any case, if the argument is for readability and easier understanding of the example rule, then all should be made consistent. There are several instances in the Rule Cookbook of using `0`/`1`, not just the single entry you modified.

ascillato:
> Timer4 should disable the rule and its missing

yes, you are right. It is a typo. Fixed. Thanks for reporting!!!!

ascillato:
> I agree, there is no timer #4 in write-up of the setup. It's likely a typo and should be time #1.

yes, exactly.

PR #535, opened 2020-11-30
Include "miscompare" error as text for easy search

plamenko:


blakadder:
great addition I overlooked. thanks!

PR #543, opened 2020-12-06
Update MQTT.md

amayii0:
Encountered MQTT connection issues and was missing return codes table.
Proposing this change as per https://github.com/arendst/Tasmota/issues/8275


ascillato:
Thanks.

PR #545, opened 2020-12-09
reverses definition of SwitchModes 1 and 2

vic42:
Reverse definition of SwitchMode1 and SwitchMode2 to reflect actual software behavior
A LOW level (0V) on a switch GPIO input actually results in {"POWER":"OFF"}
a HIGH level (3.3V) on a switch GPIO actually results in {"POWER":"ON"}.


ascillato:
Hi, thanks for your input, but it is correct as it is. 

https://github.com/arendst/Tasmota/issues/10099#issuecomment-741817295

PR #548, opened 2020-12-11
Update PWM-dimmer-switch.md

pcdiem:
I included a couple images as exmaples of non-dimmer switches that can make use of the PWM Dimmer module. I can't upload the images to the _media folder so the links are pointing to a local server. Please copy/update the images and move them to the _media folder.


barbudor:
Hi @pcdiem
Your images should be cropped (current 1476 x 1242)
There should be no reason that you can't include them in your PR by uploading them into the asset directory of your own fork.

PR #554, opened 2020-12-13
Fix markup and a typo in one sentence

cendern:
Fixed three single-quotes that should have been backticks which was causing the wrong parts of the sentence to be identified as code vs text.
Also corrected "thab" to "than".


blakadder:
thank you

PR #564, opened 2020-12-23
Update for the heltec wifi kit 32

billygr:
Includes template for the heltec wifi kit 32


billygr:
Fixes documentation issue : https://github.com/arendst/Tasmota/issues/10223

PR #567, opened 2020-12-26

MichiM2:
added definition for EasyMeter Q3A


kugelkopf123:
Sorry for this off topic stuff but you can ask your provider for a pin code to enter extended info mode. You get a better resolution and far more values as power per phase and so on. 
Have the same meter. 
But some q3a meters are a bit tricky. They don't stay in this extended mode longer than 2 min. There is a simple workaround to hold it in extended mode. If you are interested just reply..

PR #570, opened 2020-12-29
Introducing Doc Cookbook

barbudor:
Adding a "Doc Cookbook" section at the end of Contributing to show typical syntax for elements above Markdown and not well rendered in previewer.
Starting with the info/warning/example... boxes
To be continued ...


blakadder:
you mean like this https://github.com/tasmota/docs/blob/development/README.md ?

barbudor:
ashamed :(

PR #580, opened 2021-01-06
Add Sensor83 and EnergyReset time parm

curzon01:
see PR [#10423](https://github.com/arendst/Tasmota/pull/10423)


curzon01:
Add Sensor83 and NeoPool module

curzon01:
will do it on master

PR #586, opened 2021-01-08
fixed formatting of list in Homebridge.md

robertoostenveld:
the list with supported items was shown as ` ... the following Homekit device types: * Light bulb * Switch * Outlet * Motion Sensor ...` on https://tasmota.github.io/docs/Homebridge/. I am not 100% sure why that was, since a preview in the GitHub editor looked fine. But I think this should fix it.


blakadder:
thanks! space after the previous line is key due to strict markdown enforcement

PR #591, opened 2021-01-12
Fix indent by 4 spaces

jcisio:
Currently mkdocs requires that block indents are 4 spaces (even Github accepts 3 spaces). So this fix that, it works well in both Github preview and Mkdocs.

Before
Before
After
After


blakadder:
thanks

PR #595, opened 2021-01-13
 adding buttons connection & configuration schematic

barbudor:
I've used relative html links to embedd the PNG and linked the SVG
should work


barbudor:
All good now

PR #613, opened 2021-02-02
Update Home-Assistant.md

timmmmmmmmm:
Just came from a fresh Tasmota install (with no prior MQTT connection set up) and it wasn't clear to me that the MQTT settings on the device also needed to be configured.


blakadder:
already mention at the beginning of the guide 
image

PR #618, opened 2021-02-04

MubiTec:
Apator Norax 3D works with same code as Hager EHZ363 (SML)


blakadder:
thank you for your contribution

PR #633, opened 2021-02-14
Added example of another temperature monitoring device

ccoetzer:
If the device is specified incorrectly then MQTT gets spammed with messages and things appear to be working.
By specifying the device correctly polling instead happens every 70's instead of the 1-2 seconds.


blakadder:
I don't understand the point of this addition?

ccoetzer:
I had an issue where i specified the device incorrectly as just DS18B20#temperature, instead of prefixing it with tele-

 I thought this would add clarity to the docs for people using different devices.

PR #663, opened 2021-03-07
Zigbee documentation update

sanchosk:
I updated a few commands in Zigbee documentation.
My aim was to reverse-engineer the release changelog to be a bit more up to date.
Unfortunately, I don't understand all the commands and for some of the SO I cannot see a difference in output. Perhaps someone more skilled can help here...


sanchosk:
Will provide a new one with hints from @sfromis ....

PR #666, opened 2021-03-09
Dimmer < > and ! parameters

nohat:
Document the new `<`, `>`, and `!` parameters for the `Dimmer` command added for [#11269](https://github.com/arendst/Tasmota/pull/11269)


barbudor:
May I suggest a wording that'd better highlight that it is not a single step change but start a continuous change ?
May be:
> starts fade down to 1
> starts fade up to 100
If you agree you can update you branch and that will automatically update the PR.

nohat:
The [Zigbee Cluster Library specification](https://zigbeealliance.org/wp-content/uploads/2019/12/07-5123-06-zigbee-cluster-library-specification.pdf) has a Level cluster (3.10) with "Move" and "Stop" commands that support this type of behavior. The terms "move" and "stop" are here used intentionally to parallel the names for these commands from the ZCL spec, which these new commands are designed to emulate.

The ZCL is also the basis of both Dotdot and Project CHIP's command and control frameworks, and I expect it to be the de facto standard for lighting and cover level control. I covered this in substantially more detail in [the PR](https://github.com/arendst/Tasmota/pull/11269) for the related code changes.

All that being said, if the way I described the commands in this documentation update was unclear to you, it will certainly be unclear to others. I agree it should be clear that the `Dimmer <` and `Dimmer >` commands are intended to begin fades and not for instantaneous jumps, although if `Fade` is `OFF` they will be instantaneous jumps. I will update the PR accordingly.

PR #667, opened 2021-03-09
New documentation for LCD/DLP Projector control

jbubik:
This should be in section "Features"


barbudor:
@jbubik Unless I'm wrong you are adding a new page but there is no link to it neither it is added to the TOC
That looks like a dead page

jbubik:
@barbudor Yes, I didn't know how to do that. Seems mkdocs.yml is the right place. I will make a new PR.

barbudor:
It was not necessary to close
You can add more changes to your fork's branch and that will automatically update the PR

PR #671, opened 2021-03-12
Update Rules.md

nagyrobi:
Add RfRaw command duplicator example to cookbook, without specifically naming Home Assistant.

### DIFF
--- a/docs/Rules.md
+++ b/docs/Rules.md
@@ -1966,7 +1966,7 @@ The only catch is that the protocol needs to be setup in the rule. Most likely t
 
 ### RfRaw command duplicator
 
-Sonoff RF bridge with Tasmota firmware is able to send out the radio-frequency commands very quickly. If some of your covers 'miss' the commands occassionally (you can see that from the fact that the state shown in the HA system does not correspond to reality), it may be that those cover motors do not pick up the codes immediately when they are sent.
+Sonoff RF bridge with Tasmota firmware is able to send out the radio-frequency commands very quickly. If some of your RF covers 'miss' the commands occassionally (you can see that from the fact that the state shown in the HA system does not correspond to reality), it may be that those cover motors do not pick up the codes immediately when they are sent (it may happen not only with RF covers).
 
 This can be handled by sending out the RF code immediately again. While this is possible by repeating the automation sequence in HA system, it is also possible to do the repetition in Tasmota, and send the code only once from your home automation.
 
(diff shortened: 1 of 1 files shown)


docs/Rules.md
```diff
@@ -1966,3 +1964,41 @@ The only catch is that the protocol needs to be setup in the rule. Most likely t
 
 ------------------------------------------------------------------------------
 
+### RfRaw command duplicator
+
+Sonoff RF bridge with Tasmota firmware is able to send out the radio-frequency commands very quickly. If some of your covers 'miss' the commands occassionally (you can see that from the fact that the state shown in the HA system does not correspond to reality), it may be that those cover motors do not pick up the codes immediately when they are sent.
```
jaymemaurice, line 1969: This should probably be rewritten not just specific to curtain motors.

jaymemaurice: or mentioned "as an example"

jaymemaurice (changes requested):
Good work :)

docs/Rules.md
nagyrobi: The whole cookbook is a collection of examples actually, no need to mention again. It is not specific to curtain motors, but this happens most often with those, so this description was written with google search in mind, to find here the solution.

barbudor:
@nagyrobi
Have you validated your syntax by generating the pages locally (mkdocs build) ?
It's missing mandatory line feed between blocks/sections.
Please generate and confirm that the look and feel is correct.

>>> DO NOT TRUST GITHUB PREVIEW FOR MKDOCS GENERATED DOCUMENTS <<<

barbudor:
@nagyrobi 
This section need rework
Thanks

image

nagyrobi:
Thank you for your note, please check now.

barbudor:
@nagyrobi 
This is not the correct way. Please don't force `<br>` when that can be avoided.
What were missing are empty lines:
```
#### Result

The backlog coming through MQTT will:

- first set the first 32 characters of the code in variable `var1`;
- fire an event with variable `varf` containing the rest of the characters

The rule will detect the event (`on event`) and run a new backlog which will:

- run command RfRaw with code concatenated of `var1varf`;
- run command RfRaw with code concatenated of `var1varf` again;
- run command RfRaw 0

If you want to repeat the RfRaw 3 times, all you need to do is to simply add in the rule a new `RfRaw %var1%%value%;` before `RfRaw 0`.

Backlog introduces an [inter-command delay](https://tasmota.github.io/docs/Commands/#setoption34) of 200 milliseconds by default. If you feel the reaction time between your HA system and the covers is too long, you can tweak it by reducing the delay to, for example 60 milliseconds with `SetOption34 60`.
```

Insertion of HTML tags should strictly limited to when it's not possible to do otherwise, such as line break inside tables (such as in Commands table).

image

barbudor:
Thanks @nagyrobi

nagyrobi:
Any idea why it doesn't appear on https://tasmota.github.io/docs/Rules/#rule-cookbook ?

barbudor:
Because it was made against development, not master. We would need to merge it to master.

PR #674, opened 2021-03-13
New documentation for LCD/DLP Projector control

jbubik:
TOC entry added


barbudor:
@blakadder I checked that the page builds ok but may be you want to advise on the location of the entry in the TOC

PR #680, opened 2021-03-19
Started documenting new toy Inkbird IHC-200

jaymemaurice:
Started documenting new toy Inkbird IHC-200.
I don't think I will have time to finish reverse engineering it but maybe someone who needs an MQTT humidity controller will pick up where I left off.


blakadder:
thanks

PR #687, opened 2021-03-24
add documentation for configuration of spacetronik smart sl-ps26 power strip

fast-potat0:
I have configured spacetronik sl-ps26 power strip with tasmota, so I am sharing configuration.


blakadder:
please use https://templates.blakadder.com for new devices and their templates

PR #694, opened 2021-04-02
Document the SerialBuffer command

eriktews:
This documents the new SerialBuffer command and explains how it can be
used to fix problems with serial buffer overflows when using a P1 meter.


barbudor:
@eriktews 

# docs/Commands.md
The code states that the parameter can be between 256 and 520 but you wrote 510. Can you please fix?
type : `Sometimes, _serial buffer overruns_ can be minitaged by setting this to a large value` : should be _mitigated_

Please fix the typos on your fork.
Thanks

eriktews:
Has been fixed in both files and the branch has been rebased

PR #726, opened 2021-05-06
Erase flash memory

flywire:
Device does not need to power cycle when erasing flash memory with esptool.py or it's derivatives while flashing


flywire:
Note for `esptool.py` and `esptool executable` can erase memory on same command line as flashing. Both versions use exactly the same commands which should be listed once.

blakadder:
its always recommended to power cycle after erasing flash

flywire:
That's old advice: https://github.com/espressif/esptool/releases/tag/v2.6 and not compatible with Tasmotizer

blakadder:
Then change the esptool.py section not general advice.

What do you mean not compatible with Tasmotizer?

flywire:
Erasing flash can be essential before flashing but the main point is it doesn't _require_ power cycling.

I assumed Tasmotizer was running esptool.py code so the same comments would apply, especially given the following. Tasmotizer has a tick box to `Erase before flashing` but it seems Tasmotizer must flash when it runs (ie can't erase without flashing).

PR #727, opened 2021-05-06
Remove GPIO0 and GND wired bridge after flashing

flywire:
Describe the need for bridge removal after flashing. Suggested text provided,

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -145,7 +145,7 @@ To put the ESP8266 into Programming Mode:
 1. Disconnect serial-to-USB adapter and power
 2. Bridge GPIO0 and GND (by pressing the on-board button or connection with a wire)
 3. Connect the serial-to-USB adapter to your computer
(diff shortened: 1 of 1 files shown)


blakadder:
what is **run mode** ?

flywire:
If there is a `program/flash mode` there must be a mode to describe the other state. Based on my search it seems to be `run mode`.

It's only suggested consistent text for discussion with the main point being to ensure any bridge is removed after flashing.

Jason2866:
There is no `run mode`. The official modes are https://github.com/espressif/esptool/wiki/ESP8266-Boot-Mode-Selection#select-bootloader-mode

blakadder:
> It's only suggested consistent text for discussion with the main point being to ensure any bridge is removed after flashing.

And you're making that point repeatedly

flywire:
> There is no `run mode`. The official modes are https://github.com/espressif/esptool/wiki/ESP8266-Boot-Mode-Selection#select-bootloader-mode

Of course there is, `run` is a simpler term for `execute`. You are arguing there is no `program/flash mode` either.

> GPIO0 Input     | Mode
--------------- | ----------------------------------------------
Low/GND         | ROM serial bootloader for esptool.py
High/VCC        | Normal execution mode

Which eventually refers to [Expressif's docs](https://www.espressif.com/sites/default/files/documentation/ESP8266_Pin_List_0.xls):  `FLASH BOOT` and `UART BOOT`.

flywire:
> And you're making that point repeatedly

I can't argue with that logic, I included it in text repeated at the end of each flashing method.

PR #730, opened 2021-05-06
Create Azure-IoT-Central.md

ksaye:


blakadder:
what's the difference between this and #729 ?

ksaye:
729 is for IoT Hub -- which is a PaaS.  This is for IoT Central, which is a SaaS.  2 different products.

PR #739, opened 2021-05-24
Update TSL2561.md

sillyfrog:
Update to confirm `USE_TSL2561` no longer in sensors, and fix copy/paste error.

### DIFF
--- a/docs/TSL2561.md
+++ b/docs/TSL2561.md
@@ -1,10 +1,8 @@
 # TSL2561 light sensor
 
-!!! info "This feature is included only in tasmota-sensors.bin" 
-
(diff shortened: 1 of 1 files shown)


blakadder:
What's the copy paste error?

sillyfrog:
The `#ifndef USE_VEML6070` rather than `#ifndef USE_TSL2561` (I'm assuming that's what it was - but I may have missed the mark entirely). I found I needed the later (ie: as per the patch) for in my (quick) testing.

Thanks!

blakadder:
i think that was introduced because those two drivers used to clash with each other

sillyfrog:
Ahhh! Did you want me to undo that? I can add a comment to the same effect (I don't have an actual VEML6070 to test with unfortunately).

PR #753, opened 2021-06-08
Include the GUI display name where it differs

sillyfrog:
This is something I needed when searching for what `Switch_n` is in the GUI. Happy to lay it out in a different way (eg: a dedicated column, or anything else you suggest).

The goal is to have it so the name displayed in the GUI is searchable in the docs (which it currently isn't).

### DIFF
--- a/docs/Components.md
+++ b/docs/Components.md
@@ -20,231 +20,231 @@ Read more about [peripherals](Supported-Peripherals.md).
 
 ## GPIO Conversion
 
-| Old GPIO | New GPIO | Name            | Description                                                                                |
-|----------|----------|-----------------|--------------------------------------------------------------------------------------------|
-| 255      | 1        | User            | User                                                                                       |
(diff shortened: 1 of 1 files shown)


blakadder:
why not use the GUI name as it is shown?

sillyfrog:
To clarify, in the GUI, it has the option as `Switch_n` (see below) - I wanted to search for this in the docs to find out what it means. In the docs, for example, it has `Switch1n`. So when I entered `Switch_n`, I got not results. I was trying to come up with a way to make it more discoverable (hopefully that makes sense?) - happy with any other ideas :) 

Cheers.

image

blakadder:
Change the `Switch1n` to `Switch_n 1` and so forths... The former is pre 9 nomenclature while the latter is used now

sillyfrog:
How's that look?

blakadder:
almost perfect

PR #758, opened 2021-06-17
Version number fix

MDXDave:


blakadder:
thanks for catching that ;)

PR #769, opened 2021-07-01
Update Rules.md

jay-wilkinson:
Correct minor typo.


blakadder:
every typo matters! ;)

PR #772, opened 2021-07-13
Update DS18x20.md due to typo

ralf33609:
typo. tested with my own esp8266 and ds18b20 sensor. Maybe this correct has to be done for the multiple sensor example rule too. cannot verify.


blakadder:
good catch

PR #791, opened 2021-08-24
Added a warning about 4096bit RSA keys

Avamander:
When using a 4096bit RSA key without it being enabled, Tasmota will crash (every five seconds). This is very difficult for most to debug and diagnose. 

This PR adds a warning with a nice keyword discoverable on Google, which will very likely save a lot of people's time.


blakadder:
thanks

PR #794, opened 2021-08-26

anorak-47:
Added script for smart meter eBZ DD3


elajoie:
Seem the code edit is outdated with the latest dev where the serial ccp had different layout.  I tried this setyp with exactly the same meter and could not get it working.

Here is the new code edit if I did it right:
```
#ifdef ESP32
    config = (2 == m_stop_bits) ? SERIAL_8N2 : SERIAL_7E1;
#endif  // ESP32
```

PR #803, opened 2021-09-11
Elaborating on customization

barbudor:
Some elaboration

1. distinguish between 
- changing default settings that is possible on any variant
- adding/removing features that is possible only on tasmota/tasmota32

2. adding a TOC entry "Advanced customization" which include
- existing entry for multiple custom binaries
- adding description for settings that are more advanced as not necessarily matching a command such as USER_BACKLOG

### DIFF
--- a/docs/Compile-your-build.md
+++ b/docs/Compile-your-build.md
@@ -30,7 +30,7 @@ There are mainly 2 type of possible customization:
 
 - Adding or removing features. This is essentially supported only on the base **tasmota** (or **tasmota32** for ESP32). Other variants have been fine tuned and trying to add/remove features to them is most likely to fail and Tasmota development team will provide no support. The typical failure is trying to add sensors to `tasmota-display`or adding displays to `tasmota-sensors`. The proper way is to add both sensors and displays to `tasmota`.
 
-!!! Failure "Do not try to add or remove features to a variant, only to tasmota/tasmota32"
+!!! Failure "Do not try to add or remove features to a variant, only to tasmota or tasmota32"
 
 ### General customization principle
 Create a new file in `/tasmota` folder called `user_config_override.h`. You can copy the sample file `user_config_override_sample.h` that is already there and which include some sample definition for coding your own Wifi SSID and pasword inside the Tasmota firmware.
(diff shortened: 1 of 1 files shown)


ascillato (approved):

barbudor:
@ascillato Thanks for the BIG mistake forgetting the "NOT" ;)

AFAIK, no features USE_xxxx may have values
So within the block `### Enabling a feature in "tasmota"` the following doesn't seem relevant:
```
If the feature you want to customize have a value like for example:  "#define WIFI_CONFIG_TOOL  WIFI_WAIT", the best practice to modify it is to use
` ` ` c++
#ifdef %identifier%
#undef %identifier%
#endif
#define %identifier% %value%
` ` ` 
```

sfromis:
Better use correct rule syntax instead of....
`#define USER_RULE1 "On Switch1#state DO publish cmnd/otherdevice/POWER %value%"`

(Off topic, but occasionally I'm wondering if "we" have some sort of recommended "best practice" around using lower or upper case letters for rule keywords and values. While it mostly does not really matter due to the case being folded to upper, from a style perspective it can look confusing with liberal mixing. My preference is to use lower case except for things which other places is reported as upper case, like names used in JSON, but that's just me).

barbudor:
Thanks @sfromis for pointing the missing ENDON

I don't know if we have a standard style.
Myself I tend to use capital for the rule's keywords  (ON DO ENDON) and 1st capital for the rest. But even there I was not cohérent with myself :)

I'll fix that tomorrow

ascillato:
> AFAIK, no features `USE_xxxx` may have values

Yes, that is the intended standard, but there are some others, like for example: USE_AC_ZERO_CROSS_DIMMER that is defined as:
```cpp
#define USE_AC_ZERO_CROSS_DIMMER      1
```
That one is corrected in the unmerged PR 13004 in Tasmota, but as it is possible to have others too with that schema, I think is better to let that information there too. Many users just do a fast skimming of the docs instead of reading, so in order to avoid issues for new users, I think it is better to explain also that. What do you think?

ascillato:
> I'm wondering if "we" have some sort of recommended "best practice" around using lower or upper case letters for rule keywords and values.

> Myself I tend to use capital for the rule's keywords (ON DO ENDON) and 1st capital for the rest.

I also try to use it in that way. I don't know why. It seems better for reading it. Do you agree?

sfromis:
> I also try to use it in that way. I don't know why. It seems better for reading it. Do you agree?
I came to dislike all caps in computer language keywords a long time ago, aiming for a consistent style without needing any caps in keywords, and many languages refrain from treating upper and lower case the same. When I started in IT, most of the keyboards I used did not even have lower case, but newer stuff tends to "prefer" lower or mixed case.

ascillato:
I understand and I agree with you for programming languages. But if we propose a style for Rules, it should be one that makes easier to read it. When programming in VSC, the editor provides colors to commands and syntax, so it is very easy to spot commands, vars, etc., etc.

But in Tasmota rules, we don't have that color separated commands, so that is why is easer, at least for me, to spot at once the whole rule using the style of @barbudor.

For example, if using the `lua` after the 3 ticks, we have:

```lua
Rule1 ON Switch1#State DO Publish cmnd/otherdevice/Power %value% ENDON

Rule1 On Switch1#State Do Publish cmnd/otherdevice/Power %value% Endon

rule1 on switch1#state do publish cmnd/otherdevice/power %value% endon
```

But if we use `js` after the 3 ticks, we have:

```js
Rule1 ON Switch1#State DO Publish cmnd/otherdevice/Power %value% ENDON

Rule1 On Switch1#State Do Publish cmnd/otherdevice/Power %value% Endon

rule1 on switch1#state do publish cmnd/otherdevice/power %value% endon
```

So, in that case, Barburdor's Rules Coding style is good when using JS colors. What do you think?

sfromis:
> good when using JS colors. What do you think?

Due to a major point of color styling being spotting errors, I only find it useful if language-aware. And what Discord/Github may do is not a consideration for me. 

Not language-aware:
```js
Rule1 OM Switch1#State DU Publish Cmnd/otherdevice/Power %value% ENDDO
```
But whatever....

Jason2866:
My 2 cents. Documentation is not programming. So Highlighting or Coloring to make the readability better is good. Perfect if we reach a common formatting style.
@barbudor @ascillato thx for the work and pointing out.

PR #806, opened 2021-09-14
Update Rules.md

akseidel:
Conditional Rules is a plural expression but a singular is used to refer to "them" in the specific file alteration instructions. This is confusing.


blakadder:
this is against standard scheme used across the docs for features not supported. The "if" refers to the "feature" which is not included in the precompiled binary

PR #814, opened 2021-09-22
Fixed RF-Transceiver typos.

OnizukaX:
Changed "transciever" for "transceiver" which is the correct spelling.


blakadder:
thanks

PR #837, opened 2021-10-31
Update Serial-to-TCP-Bridge.md

duhow:


duhow:
This service keyword "ser2net" is missing, by adding it should help to find this utility in the project.

PR #839, opened 2021-10-31
Update description of SwitchMode 3–6

jonathanschneider:
These modes evaluate either a rising or a falling edge. The respectively other edge has no effect but the description suggests otherwise.

### DIFF
--- a/docs/Buttons-and-Switches.md
+++ b/docs/Buttons-and-Switches.md
@@ -69,22 +69,22 @@ SwitchMode, as the name implies, applies _**ONLY**_ to GPIO configured in Tasmot
         When connecting a momentary switch (i.e., a push-button) you will want to use `SwitchMode 3..7`.
 
 **`SwitchMode 3`**
-:    Set inverted push-button mode (`1 = TOGGLE`)
+:    Set inverted push-button mode (`0→1 = TOGGLE`)
 
     Tasmota will send a `TOGGLE` command when the button is pressed (closing the circuit). When the button is released (opening the circuit) nothing will happen. Default state is OFF and when pressed it's ON. (This trigger is known as [rising-edge](https://en.wikipedia.org/wiki/Signal_edge))
 
 **`SwitchMode 4`**
-:    Set push-button mode (`0 = TOGGLE`)
+:    Set push-button mode (`1→0 = TOGGLE`)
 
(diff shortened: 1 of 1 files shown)


stefanbode (changes requested):
I would not change because the default behavior has to be part of the docu

jonathanschneider:
@stefanbode this is starting a whole nother discussion but the state of the button in the resting or default position depends on the circuit. Saying that it's 1 or 0 is only true in some cases. For example, the signal of a button connected to Vcc is 0 in the resting or default position and with SwitchMode3 pressing it would toggle the relay (as described in the documentation). Whereas a button connected to GND is 1 in the resting or default position and with the same switch mode releasing the button would toggle the relay (contrary to the documentation). And this is only if the button is normally open – normally closed buttons make everything work the other way around again.
So I find any description of position of the button or the state of the circuit is futile as it only matches some situations. 
However, starting this discussion wasn't my intention with this PR and I am fine with describing the scenarios each switch mode was intended to. What I believe must be changed though are descriptions like `0 = OFF` or `1 = ON` in these switch modes as these signals do not switch the relay on or off. Only the rising edge (for SwitchMode3 and 6) and the falling edge (for SwitchMode4 and 5) have an effect on the relay.

ascillato:
Hi,

The logic state ( 1 or 0 ) should be in the docs. It is necessary.
The electrical state that defines if a 1 is VCC or GND is when you set the GPIO as `switch` or `switch_i`
With `Switch` VCC is 1 and GND is 0
With `Switch_i` ( i means inverted ) VCC is 0 and GND is 1.

So, I think the logic state should remain in the docs and we can add the relationship with the electrical state as a note if you want.

jonathanschneider:
> So, I think the logic state should remain in the docs and we can add the relationship with the electrical state as a note if you want.

Of course the logic state must remain the docs – no question about it. The description of the four switch modes that only evaluate edges isn't clear.

> The logic state ( 1 or 0 ) should be in the docs. It is necessary. The electrical state that defines if a 1 is VCC or GND is when you set the GPIO as `switch` or `switch_i` With `Switch` VCC is 1 and GND is 0 With `Switch_i` ( i means inverted ) VCC is 0 and GND is 1.

`Switch_i` doesn't exist in the pre-compiled release binaries. Is that a special option? And as far as I can tell `Switch_n` doesn't change the logic state: VCC is 1, GND is 0 independent of the circuit or the configuration.

barbudor:
Indeed there is no `Switch_i`. Switchmode is used to managed iverted switches

stefanbode:
Thanks for the comment. The last sentence is important and I see here also room for improvement. Let me think about a better definition

PR #841, opened 2021-11-01
Additional example for Rules#Timer

StoyanDimitrov:
Add example clarifying the usage.


blakadder:
i don't get it... why two identical examples?

StoyanDimitrov:
Because one example is no enough to convey the connection between the two 1's. What exactly those 1's mean? Where they come from? Are they bound to the Rule1? The two examples combined make the connection clearer.

barbudor:
This is breaking the presentation
why not using a generic approach such as
`Rules#Timer=<x>` when countdown `RuleTimer<x>` expires (where `<x>` = `1..8`)

StoyanDimitrov:
Now feels close to the Commands. Thank you fro the suggestion.

barbudor:
Thanks @StoyanDimitrov

PR #844, opened 2021-11-02
Append Info about Multi-Zone Heating Controller

Beormund:
Appended info about a multi-zone heating controller project along with link to repo.


blakadder:
very nice

PR #852, opened 2021-11-09
Fix enabling timers in example commands

flecmart:
Replace "Arm" attributes with "Enable" in timer commands.
I stumbled upon these example commands and wondered why my timers did not work as intended. 
I guess the "Arm" attribute was renamed at some point earlier.
Please review and eventually merge :-)


blakadder:
thanks

PR #856, opened 2021-11-22
Update Pinouts.md

pmknowles:
User has misread pin position from screen on underside photo of the PSF-B85 so clarified the pin positions


blakadder:
you've changed two images

PR #857, opened 2021-11-22
added notes for servo motors

Gifford47:
added notes for low angles movements and pwmendonstop


stefanbode:
Why does the angle change if you change the openduration. This is not intended. The rest is ok. Pwmmin and pwmmax must always be the same in any config

stefanbode:
For the pwmendstop the desired method to avoid small movement is to disconnect the power that comes from relay 1

PR #869, opened 2021-12-05
Grammar correction

pkkrusty:
Changed wording to reflect correct usage of the word 'myriad'


blakadder:
is it the correct usage? reference?

pkkrusty:
87DEB876-546B-4F6F-94A5-37D50CFFC4F5

as an adjective, does not include a/of. I recall my 12th grade English teacher beating it into us because most people use it incorrectly.

blakadder:
including me! 

grateful and humbled i learned this only today

PR #871, opened 2021-12-06
Listed fourth option for ufstype command

pkkrusty:
As defined in xdrv_50_filesystem.ino, ufs_type variable has four returned values.


blakadder:
thanks

PR #888, opened 2022-01-05
Adding a note about usage of "="

rtek1000:
Added note on use of '=' sign in explanations


barbudor:
Do we really need that ? @blakadder @ascillato ?

ascillato:
I think it is good and useful to have more clarifications about the syntax. I agree to merge this.

PR #897, opened 2022-01-13

patrickschwartz:
Added script for eBZ DD3 Meter SM-Type with SML binary hex decoding settings


blakadder:
thanks

PR #900, opened 2022-01-16
Move Wifi 2..4 to ESP8266

barbudor:


ascillato:
Thanks. This indeed was needed. :+1

PR #903, opened 2022-01-18
Update documentation for EQ3-TRV

obrain17:
This update documentation is based on the experience with the current version of the driver:
All commands and fields of the response are descibed as well as the behavior of the TRV in the different modes.

It already includes fields intoduced with:  BLE_EQ3: Fix and enhance the JSON response  [14485](https://github.com/arendst/Tasmota/pull/14485) 

 I have preserved the anchors (e.g. `<a class="cmnd" id="devlist">`), so references would still work.

Hopefully the markdown format fits to mkdocs material.
I have tested it with my own "mkdocs serve", and it is looking good.

### DIFF
--- a/docs/EQ3-TRV.md
+++ b/docs/EQ3-TRV.md
@@ -118,8 +118,8 @@ the valve:
   "state": "unlocked",
   "battery": "GOOD",
   "holidayend": "00-00-00 00:00",
-  "windowdur": 15,
   "windowtemp": 12.0,
+  "windowdur": 15,
   "day": 21.0,
   "night": 17.0,
   "offset": 0.0
@@ -174,8 +174,8 @@ added:
 | state         | child lock enabled (disables the buttons on the TRV): locked / unlocked                                                                           |
 | battery       | battery status of the TRV: GOOD / LOW                                                                                                             |
 | holidayend    | end of holiday mode                                                                                                                               |
-| windowdur     | window open duration                                                                                                                              |
 | windowtemp    | window open temperature                                                                                                                           |
+| windowdur     | window open duration                                                                                                                              |
 | day           | comfort temperature                                                                                                                               |
 | night         | reduction temperature                                                                                                                             |
 | offset        | offset temperature                                                                                                                                |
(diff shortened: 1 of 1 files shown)


barbudor:
Thanks
Holded until Tasmota PR #14485 is confirmed

barbudor:
Looking at the diff, this looks like a massive change (total rewrite) as not a single line of the original doc remains.
Are you sure that was necessary ?
Or is it a bad operation ?

blakadder:
Impossible to track what changed, please revise by including only the new changes

obrain17:
Basically only some parts are a rewrite. As I wrote, based on my tests and experience I made with this excellent driver.  

- However, the only part referring to PR 11485 is the description of: windowdur/windowtemp/day/night/offset in the JSON formatted response.
- All other has already been with the EQ3 driver as it exists. I only tried to descibe them more in detail

There are too many changes and additions to write them manually into the md. So I used my own editor that produces a different syntax for some markdown elements. 
That's why a diff is not possible: just load both in mkdocs and compare side-by-side.

Originally I intended to do this for my personal documentation only. But then I thought to make it publicly available, added the anchors etc. to make it behaving and looking like the orignal.

barbudor:
@obrain17
Please clean-up the PR to only include changes as discussed
Thanks

obrain17:
This is already done:
The PR includes only changes that are needed to describe the driver's actual behavior. 

Many lines have been unchanged, but "Diff" unfortunately shows all lines as changed.

A better comparision can be done side by side either with mkdocs or with notepad++ compare.

obrain17:
I now have manually edited the file:
With some modifications in the markdown syntax, "diff" now shows the changes.
The web presentation with mkdocs is the same of course.

barbudor:
Just for info, you don't need to change the numbers in a numbered list. Just leave it a `1.` for every line and mkdocs will do the job. That really save times when editing such lists.

@blakadder Any other comment ?

blakadder:
No clue about that driver so I have no say

PR #904, opened 2022-01-18
Update BlitzWolf-SHP6.md

shafr:
Added comment on different pinout. Checked in 2 sockets & confirmed with thingiverse programmer schematic:
https://www.thingiverse.com/thing:4093593/files


barbudor:
Hi @shafr 
Would you happen to have a photo for this alternative pinout ?
A way to distinguish such as some marking on the PCB ?
Thanks for sharing

shafr:
Just those one from adapter on thingiverse. 
or this might work ?

SHP6_v2_contacts_upd

PR #907, opened 2022-01-21
Describe AC switch adapter usage in Shelly-RGBW2.md

matrach:
It took me a while to notice, that enabling the internal pull-up breaks edge detection on the AC adapter connected to the SWAC input, since such setup worked with an oscilloscope. Probably, there is some special circuitry on the PCB. 

I'm attaching photos of the PCB for reference.


barbudor:
thanks

PR #912, opened 2022-01-25
Update for adc

francescoadriani:
Mq on guide, image with all adc input and image for esp32;
Update for adcparam


barbudor:
@francescoadriani 
Is it how it is supposed to look like ? 
Shouldn't there be a space where I highlighted in red ?
(just wondering)

image

francescoadriani:
It is const defined on source. Could be everything on user manual. Moreover it isn't used yet! How do you prefer?

barbudor:
OK, I was just wondering if it was what is intended.
Thanks for the collaboration

francescoadriani:
Yes it is ppm supposed for clean air! Thanks to you to support us

PR #916, opened 2022-01-28
Add EMH ED300S to Smart Meter Descriptor examples

emcell:
Adds the EMH ED300S (SML) to the Smart Meter Descriptor examples.


barbudor:
Thanks

PR #917, opened 2022-01-29
I2C devices in simpler english

fabian727:
I had my problems reading the first paragraph and was confused that I2C is enabled in the official releases, but not on custom builds. It was also not enabled by setting USE_MCP230xx

There was a copy paste error in the used devices. I use my MCP23017 with address 0x27

### DIFF
--- a/docs/I2CDEVICES.md
+++ b/docs/I2CDEVICES.md
@@ -42,8 +42,8 @@ Index | Define              | Driver  | Device   | Address(es) | Description
   19  | USE_SI1145          | xsns_24 | SI1147   | 0x60        | Ultra violet index and light sensor
   20  | USE_LM75AD          | xsns_26 | LM75AD   | 0x48 - 0x4F | Temperature sensor
   21  | USE_APDS9960        | xsns_27 | APDS9960 | 0x39        | Proximity ambient light RGB and gesture sensor
-  22  | USE_MCP230xx        | xsns_29 | MCP23008 | 0x20 - 0x27 |  8-bit I/O expander
-  22  | USE_MCP230xx        | xsns_29 | MCP23017 | 0x20 - 0x27 | 16-bit I/O expander
+  22  | USE_MCP230xx        | xsns_29 | MCP23008 | 0x20 - 0x26 |  8-bit I/O expander
+  22  | USE_MCP230xx        | xsns_29 | MCP23017 | 0x20 - 0x26 | 16-bit I/O expander
   23  | USE_MPR121          | xsns_30 | MPR121   | 0x5A - 0x5D | Proximity capacitive touch sensor
   24  | USE_CCS811          | xsns_31 | CCS811   | 0x5A        | Gas (TVOC) and air quality sensor
   25  | USE_MPU6050         | xsns_32 | MPU6050  | 0x68 - 0x69 | 3-axis gyroscope and temperature sensor
--- a/docs/I2CDEVICES.md
+++ b/docs/I2CDEVICES.md
@@ -1,7 +1,7 @@
 
 # I2C devices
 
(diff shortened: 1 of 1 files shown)


docs/I2CDEVICES.md
```diff
@@ -42,8 +42,8 @@ Index | Define              | Driver  | Device   | Address(es) | Description
   19  | USE_SI1145          | xsns_24 | SI1147   | 0x60        | Ultra violet index and light sensor
   20  | USE_LM75AD          | xsns_26 | LM75AD   | 0x48 - 0x4F | Temperature sensor
   21  | USE_APDS9960        | xsns_27 | APDS9960 | 0x39        | Proximity ambient light RGB and gesture sensor
-  22  | USE_MCP230xx        | xsns_29 | MCP23008 | 0x20 - 0x26 | 16-bit I/O expander
-  22  | USE_MCP230xx        | xsns_29 | MCP23017 | 0x20 - 0x26 | 16-bit I/O expander
+  22  | USE_MCP230xx        | xsns_29 | MCP23008 | 0x20 - 0x27 |  8-bit I/O expander
+  22  | USE_MCP230xx        | xsns_29 | MCP23017 | 0x20 - 0x27 | 16-bit I/O expander
```
barbudor, line 46: There is a good reason that 0x27 is NOT listed
0x27 has been excluded to avoid conflict with some other devices such as the I2C LCD display
From the driver's code:
```
/*********************************************************************************************\
   MCP23008/17 - I2C GPIO EXPANDER

   Docs at https://www.microchip.com/wwwproducts/en/MCP23008
           https://www.microchip.com/wwwproducts/en/MCP23017

   I2C Address: 0x20 - 0x26 (0x27 is not supported)
\*********************************************************************************************/
```

barbudor:
> I2C is enabled in the official releases, but not on custom build
This doesn't makes sense
I2C is enabled in `tasmota`/ `tasmota32` which is the base for custom builds.

fabian727:
I flashed the official 10.1 release on a wemos d1 mini.
The [mcp23017](https://www.ebay.de/itm/255313705406) is set with pullup resistors for a0-a2. The board is connected to GND, 3.3V D4 and D3 for SDA and CLK. Reboot and an I2Cscan ffound the MCP23017 at 0x27.
A simple "sensor29" written in the online console returned error, as I didn't know about the correct usage eg "sensor29 1,1,1".
I used tasmotizer, set to generic (though I am using above mentioned wemos d1 mini)

`#define USE_MCP230xx                 // Enable INPUT mode (pinmode 1 through 4)
#define USE_MCP230xx_OUTPUT          // Enable OUTPUT mode (pinmode 5)
#define USE_MCP230xx_DISPLAYOUTPUT   // Display state of OUTPUT pins on main Tasmota web interface
#define USE_MCP230xx_ADDR 0x27`
[copied and changed address](https://tasmota.github.io/docs/MCP230xx/#usage-of-the-driver)
recompiled, flashed and above configured pins were displayed empty. I recompiled with additional flag
´#define USE_I2C' reflashed and it got recognized again.

Therefore I guessed that this flag is needed on custom builds. Maybe it was a compile time bug or it is a real bug in the code.

docs/I2CDEVICES.md
fabian727: but it is supported and therefore wrong to state range from 0x20 - 0x26. Write a comment to this table or reference to MCP page; but it is confusing to write different ranges on different pages.

barbudor: Working doesn't mean supported.
Someone in Tasmota team decided that this value should be excluded to avoid conflicts with other device.
Please remove this change from your PR

fabian727: changed those lines back

blakadder:
this file will be overwritten by the original from arendst/Tasmota anyway so make the changes there

PR #919, opened 2022-01-30
[Solax X1] typo found and fixed

SteWers:


barbudor:
By the way, reading the text, I don't really agree that only 2 lines are needed.
For me, You still need a ground reference between the 2 circuits otherwise you may end up with a common-mode voltage between the 2 circuits that could be dangerous.
Usage of a differential signal is mainly immunity to noise on very long distances.

SteWers:
@barbudor The RS485 is designed as a 2 wire connection. It's nearly the same technology as old telegraphs. Also in modern times POTS an DSL work with only 2 wires. In my setup with a X1 and a Dual R2 it is running since weeks without any problem.
If you're interested in details, you may read this: https://en.wikipedia.org/wiki/RS-485

barbudor:
@SteWers
Thanks, I've been usingRS485 in industrial environment for many years and know quite a bit about the pifalls.

RS485 supports a maximum common mode of -7V/+12V
If you like Wikipedia, you can read again the section starting with "In addition to the A and B connections". ANd if you want to learn more, section 11 of Texas Instruments [RS-485 Design Guide](https://www.ti.com/lit/pdf/slla272) or their dedicated [training](https://e2e.ti.com/administrators1/industrial_interface/m/videos__files/264949/download) on that topic advising to use galvanic isolation on long lines where a large ground potential difference may exist between the 2 sites.
Before connecting 2 devices with RS-485, it is common sense to check if a there is a common mode voltage between the 2 devices.  Also as stated in the App Note, it is not because there is no GPD on the day of installation that may not come later if the installation is reworked.

PR #941, opened 2022-02-28
Automatically generated table of timezone commands

cpfair:
I thought it would be convenient to have a table of ready-to-use timezone configuration commands that people can copy-paste from when setting up their devices.

The generated markdown file looks like this: https://gist.github.com/cpfair/253d4fb5bab9b6d58e0a204a78003118


ascillato:
Wow !

So cool !

barbudor:
Great

Sorry for my 1st message, I didn' saw you actually contributed the script to generate it and not the static page.
Excellent job.
Many thanks

barbudor:
A noob question : what is specifc with Groenland or Israel timezone rules that it cannot be supported by Tasmota ?

cpfair:
There are a few different reasons timezones aren't supported:

* (most common) DST rules involving an offset, e.g. Tasmota supports "the first Sunday in November," but not "the Friday _before_ the first Sunday in November." This is the restriction affecting your two examples.
* UTC offsets outside of the [-13,13] hour range which Tasmota accepts (`Pacific/Kiritimati` is +14).
* DST changeover times that are not whole hours (`Pacific/Chatham` changes over at 2:45AM and 3:45AM).
* DST rules that are not based on the Gregorian calendar (`Asia/Tehran`).

I haven't looked into it much, but I think the first 3 cases could be supported by extending the range/precision of existing configuration fields. Only the last case would remain - which even the IANA timezone database [does not attempt to represent algorithmically](https://github.com/eggert/tz/blob/main/asia#L1534).

barbudor:
Thanks
As there is a script involved, I think it's better to have @blakadder approval on this PR

ascillato:
@blakadder Hi, what do you think of this PR? Ready to be merged?

PR #945, opened 2022-03-09
Add troubleshooting section for enabling and disabling I2C drivers

woutervddn:
An update as to reflect the discussion on https://github.com/arendst/Tasmota/issues/15072


barbudor:
While the text makes sense, I think is it is not well located in the AHT driver section because its use case is more general.
It would be better placed in the I2CDevices documents
The I2CDEVICE page used in the doc is a copy of the one from the tasmota repository here: https://github.com/arendst/Tasmota/blob/development/I2CDEVICES.md
Any PR to that doc pages should be done in the Tasmota repo.

A link from the AHT page (and from any other I2C specific page) to the I2CDEVICES page would also make sense if not yet existing.

Thanks

PR #952, opened 2022-03-17
Change RST to EN pin for nodemcu

psbankar:
It is EN pin which should be bridged to GND for disabling onboard chip


sfromis:
Why do you say "should"? IME, it works just fine (just tested again) bridging GND and RST. Of course, It is just fine to do it with EN too, but that does not make it an "wrong" to use RST. The point is to keep the ESP8266 in hibernation, and both bits works for that.

When I put the tip into the docs, the reason why I selected to use RST is that it is a bit simpler to connect neighbor pins, like using a standard two-pin jumper, without wires. 

Maybe amend the docs to say "bridging GND to the RST or EN pin"?

psbankar:
Oh is it? I tried it yesterday referring to these docs but RST pin did not work for me and most of the other docs/tutorials mentioned EN pin only and it worked for me perfectly so I thought it was some mistake. If both pins are fine then I can change it to EN or RST.

PR #953, opened 2022-03-19

iot-sle:
Add section about M-Bus for IR connector of ABB B21-112. Depends on the bcd changes done by gemu on sml interface.


iot-sle:
The bcd decoding is already part of development now. I could notice the merge is not done yet. At the same time it seems that the documentation of smart-meter-interfaced.md is corrupted at it's very end. Could you please take a look. I added this part directly bedind the ABB Modbus example.

PR #965, opened 2022-04-05
Embed esp_web_flasher

Staars:
Fetches all data from Tasmota's /install -repository and is configured via the manifest files from that repo.

Hope it still works outside of my test environments.


blakadder:
flasher works fine. the dropdown is a bit tiny compared to the rest of the page

the screen looks like this when using unsupported browser which is repeating info
image

Staars:
Yes, that is unnecessary repetition.

I could have sworn, that this internal state was only reached in my test platforms, if the last sentence did not show up. 🤔
If I can not reproduce this behavior,  I can delete the unnecessary first phrase.

I am totally open for design optimizations.

PR #968, opened 2022-04-13
Don't break CI: Automatically generated table of timezone commands

cpfair:
@blakadder I saw that #941 caused issues with the CI build. This PR fixes the problem, a difference in `zoneinfo.available_timezones()` when run on Mac vs. Linux - I had only tested it on Mac previously. I've confirmed it works using my own CircleCI account - sorry for the trouble with the earlier PR.


ascillato:
Hi,

Thanks for fixing it so fast. Amazing job!!!!!!

One question.

I have been testing some of the timezones and I found one that don't match.

In the table https://gist.github.com/cpfair/253d4fb5bab9b6d58e0a204a78003118 it shows for Europe/Brussels:

```js
Backlog0 Timezone 99; TimeStd 0,0,10,1,3,60; TimeDst 0,0,3,1,2,0
```

But according to https://24timezones.com/Brussels/time, it should be:

```js
Backlog0 Timezone 99;TimeStd 0,0,10,1,3,60;TimeDst 0,0,3,1,2,120
```

For the case of Europe/Brussels, STD timezone should be +1 (60min=OK), and DST timezone should be +2 (but shows 0 instead of 120).

Please, can you take a look to it?

Thanks!!!!!!!

cpfair:
Good catch - I've fixed the problems and added a number of automated tests.

ascillato:
Thanks !!!!

ascillato:
@blakadder 

Please, can you take a look if it can be merged now?

ascillato:
@blakadder

Please, can you take a look if this PR can be merged now?

blakadder:
It's breaking CI again https://github.com/tasmota/docs/runs/8155221490?check_suite_focus=true therefore it is removed

cpfair:
Perhaps a static copy of the file is a better approach: #1032

PR #969, opened 2022-04-15
FAQ: added Device unavailable if it cannot reach NTP server

Heisenberg2980:
Added comment about device being unavailable if it cannot reach NTP server to update time

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -121,6 +121,11 @@ You have more than one device connected with the same %topic% defined. Its impor
 
 If that is not the issue, erase all flash using esptool.py or esptool.exe and flash again by wire (as explained [here](Esptool#upload-tasmota)) using [the latest precompiled bins with core v2.7.1](http://ota.tasmota.com/tasmota/).
 
+### Device is often unavailable, ussually 30 seconds every minute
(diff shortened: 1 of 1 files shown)


ascillato:
Hi, this is not a normal behavior to document as such. This is a bug from the Arduino core that have been fixed twice already but it have come up again. It is reported in Tasmota at https://github.com/arendst/Tasmota/issues/14394

Heisenberg2980:
Actually https://github.com/arendst/Tasmota/issues/14394 seems to be a different issue where the physical buttons stop working when wifi is down (it is not talking about NTP/time).

This problem with NTP servers seems to be affecting a lot of people, and I think it should be documented in the FAQ section so people don´t need to search between hundreds of issues to find the cause and the possible workarounds while a solution for this problem is found.

Wouldn´t be better to add this to the FAQ section indicating the reason, and in case this is fixed at some point the FAQ section is removed?

ascillato:
Yes, you are right. Also, we should add that it is a known bug we are working on to solve.

Heisenberg2980:
@ascillato thanks for the suggestion, I have updated the comment

Heisenberg2980:
@ascillato do you have permissions to merge this PR? if not, do you know who can merge it?

PR #971, opened 2022-04-16
Fix incorrect formatting of commands page

spuder:
The mkdoc admonition format is incorrect on [this page](https://tasmota.github.io/docs/Commands/)

The extra `:` on the admonition is confusing mkdoc, This _should_ fix it (though I don't have a way to test it)

Screen Shot 2022-04-16 at 10 37 34 AM


blakadder:
thanks for the fix

spuder:
Confirmed fixed

Screen Shot 2022-04-20 at 3 58 35 PM

PR #995, opened 2022-06-10
Update example for Wifi watchdog to working state

justintime:
The current example for the Wifi Watchdog has two problems: one, it requires Ping which isn't included by default; and two, it doesn't actually work.  The `Ping4` command needs to be wrapped in a backlog in order for the ruleset to work properly.  However, we have something more flexible than Ping - WebQuery.  It's also included in Tasmota by default.  This documentation change updates the example to use WebQuery in a way that actually works.

### DIFF
--- a/docs/Rules.md
+++ b/docs/Rules.md
@@ -1761,14 +1761,18 @@ Notice we use `Rule` which edits `Rule1` rule set. They can be used interchangea
 
 ### Watchdog for Wi-Fi router or modem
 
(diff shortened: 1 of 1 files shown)


barbudor:
Maybe but you can't always assume that there is a web server available on the IP tested
So both example have their own advantage and should remain (with fixed version for ping)

justintime:
That makes sense.  I modified it to show both methods.

barbudor:
Thanks

PR #999, opened 2022-06-25
Delete OpenHASP.md

sfromis:
Deleting orphan page after Tasmota "OpenHASP" was renamed to "HASPmota" and a new page was created. I'm assuming that this is OK, or does more need to be done?


sfromis:
Superseded by s-hadinger doing the deletion with redirect

PR #1000, opened 2022-06-28
Update docs for smart meter AS2020

KG3RK3N:
Added the necessary config for the smart meter AS2020 to fix missing sign on current power value.


blakadder:
what is going on with this?

PR #1036, opened 2022-09-07
Update pin layout of THR316D/THR320D

pluk77:
Based on https://github.com/arendst/Tasmota/issues/15856 Pin layout was incorrect according to @Longlife4ever


blakadder:
thanks

PR #1043, opened 2022-09-16
Added tools section

Gifford47:
added "Tools" sub-section under help section and inserted some projects.


jeroenst:
I think this is a good addition to the documentation.

jeroenst (approved):

jeroenst (approved):

barbudor:
Most of them are already documented in another place
What I don't like in dupplicating information is how it becomes wrong when only one side is updated

blakadder:
image

PR #1048, opened 2022-09-17
Update Commands.md

erxbout:
update 
EnergyToday
EnergyYesterday
EnergyTotal
(summing up today and yesterday because the description is the same)

remove typo (At least I guess Set or set is not intended..)
add warning sign to depricated command EnergyReset so one can spot it easier


blakadder:
do not merge commands since the anchors are used elsewhere

erxbout:
If I edit it like this, the anchors are still accessible with the same ids
Is that ok?

jeroenst:
I think this change is working already.

When browsing to the following urls everything seems fine:
https://github.com/tasmota/docs/blob/master/docs/Commands.md#energytoday
https://github.com/tasmota/docs/blob/master/docs/Commands.md#energyyesterday
https://github.com/tasmota/docs/blob/master/docs/Commands.md#energytotal

jeroenst:
I think this addition is ok, but don't change the anchors indeed, they work as they are intended.

jeroenst:
I created a new pull request that solves this issue reported without changing anchors.

erxbout:
I do not quite understand yet.. can you explain what do you mean by "not change anchors"
In my first commit I changed them, but in the second reverted them and used them with the ids from before, just in the same row..

Two commands in the same row should then work or am I missing something?
@jeroenst also you missed my change regarding EnergyReset.. I added a warning there so it will be seen better in future..

and also there is a reference to "the commands above" but one of them (EnergyYesterday) is below and one of them is above.. that also adds some confusion..

barbudor:
If this is no more needed could you please close this PR ?
Thanks

jeroenst:
@erxbout Ok, done.

erxbout:
I meant a warning sign like this :warning: 
or does this only work on github..?

Do not want to sound argumentative, just want to understand.. 
But it feels like I am not able to apparently.. :(
anyway have a nice day..

jeroenst:
I don't know if that works in the documentation builder. Anyway, thank you for your input!
I only tried to help.

barbudor:
I think hte problem raised by @blakadder was that part:
`EnergyToday<x\><a class="cmnd" id="energytoday"></a><br>EnergyYesterday<x\><a class="cmnd" id="energyyesterday">`
as you have merged 2 anchors into the same block of the table

Regarding the warning sign, I'm not sure this is working. I would be more tempted to use a Unicode

Don't forget that what you is in Github markdown preview can be very different than wat is actually rendered by mkdocs
If you have any doiubt, the best is to clone loclly, install the mkdocs tool chain and generate the docs locally

erxbout:
I wanted to understand why, so I needed to test it myself.. 
this might be interesting but idk
Indeed there is strange behavior if you merge commands in one row see screenshot here:

2022-09-26-232548_1600x900_scrot

Seems like the anchor needs more space when going there by clicking an anchor ref, also when accessing energyyesterday it points to the command above instead of energytoday..
What you can also see is that the warning sign (as already used command description for Reset for example) is working correctly

there would have been no need for a new pr, I could have changed these things but I wanted to understand them, now I do

PR #1049, opened 2022-09-19
Add MI32 legacy page

Staars:
This is intended to be an unlinked page as this driver is not part of any standard build.


blakadder:
fine by me

PR #1055, opened 2022-09-25
Add information about SetOption147

ortegafernando:
Be carefull about the missing options (from 136 to 146). Thanks.


barbudor:
thanks

PR #1058, opened 2022-09-26
Update MQTT.md

jeroenst:
Added stateretain, inforetain and statusretain


barbudor:
Same, AFAIK this is not available in Master so it should be added to Development

jeroenst:
Created new PR based on development branch

PR #1064, opened 2022-09-27
Update Sensor54 commands

pkkrusty:
From #16653


barbudor:
If it works for channel 1 (Sensor54 13), although not tested, it should work the same for channels 2 and 3 (Sensor54 23 and Sensor54 33)

barbudor:
I made the changes
Good to merge now
Thanks

PR #1066, opened 2022-10-01

ralf-koenig:
Better separation of the M-Bus example in the ABB section. As this is the only M-Bus example on this page and the only example using "r" (raw mode), it should stand out better and be clearly distinguishable from the Modbus example on the same meter. This is especially true as M-Bus and Modbus could otherwise easily be mistaken.


blakadder:
thanks

PR #1068, opened 2022-10-02
Fix markdown link in readme

flabbamann:
Old link returns 404...


blakadder:
since the original link seems to be changing on a steady rotation i replaced it

barbudor:
Sorry @blakadder

PR #1074, opened 2022-10-08
fix indentation for Logarex LK13BE (SML)

abrychcy:
Script was not properly displayed in documentation. This commit fixes the issue.


blakadder:
thanks

PR #1076, opened 2022-10-12
Update DS18x20.md

intsoftl:
Clarification to ## THR316D/THR320D Wiring Notes following implementing on THR316D

The plug is rj10 4p4c , not rj11.
Note added that no pullup resistor needed as included in hardware already.


blakadder:
thanks

PR #1094, opened 2022-10-28
Update Rules.md

jsiddall:
Added note about precision when using arithmetic commands used with VAR

### DIFF
--- a/docs/Rules.md
+++ b/docs/Rules.md
@@ -1353,7 +1353,9 @@ Rule2
 ------------------------------------------------------------------------------
 
 ### Arithmetic commands used with VAR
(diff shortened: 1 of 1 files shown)


docs/Rules.md
```diff
@@ -1353,6 +1353,7 @@ Rule2
 ------------------------------------------------------------------------------
 
 ### Arithmetic commands used with VAR
+Note: arithmetic is done using single point precision floating point.  This means calculations involving values larger than approx. 16 million (ex: `%utctime%`) will not be precise.
```
barbudor, line 1356: Correct english would be "will not be accurate"
Thanks

jsiddall: In general English, yes, but in the context of math precise and accurate are different things.  In this case the result _will_ be accurate, just not precise.

PR #1095, opened 2022-10-30
New alternate version of daylight timer

cybermaus:
Using %timerN% variables, settable in Web-UI, and works past midnight and in high north.


cybermaus:
@blakadder  not wanting to disturb anyone, but I am suddenly wondering if this is not visible because of the "workflow approval" message. Is there something I need to be doing to resolve that? or am I just impatient?

blakadder:
you're wondering too much, relax

PR #1103, opened 2022-11-05
Update DS18x20 docs

krystof-k:
I updated the docs with regards to the recent DS18x20 enhancements  introduced in arendst/Tasmota#16649 and arendst/Tasmota#16833, ready for the v12.3 release.


krystof-k:
@md5sum-as, I would be happy if you can check it :)

md5sum-as:
Hello.
Everything is right.
Should we specify that when n=0 (in the DS18 command Alias \<sensor id\>,\<n\>) the alias will be deleted and the default schema will be returned?

krystof-k:
Sure! Added.

blakadder:
thanks

PR #1111, opened 2022-11-15
Update Flash-Sonoff-using-Raspberry-Pi.md

gunnaraas:
These docs present instructions for flashing using a Pi 3 as the default, but terminal output shows flashing over /dev/ttyAMA0. Updated terminal output for consistency. 

Also added reference to Pi 4, as that is the model most end users looking at the doc are likely to be using today and moving forward.


blakadder:
Thanks

PR #1125, opened 2022-12-07
Update ESP32.md to include SetOption146 and Warning to not use it 

alek67890:
After a new release CPU temp disappear and at first I was searching in docs. Later I found out about change in Changelog, so I think it should be also here.

I don't know what about, "SetSensor127 0", so I replaced it for new command

### DIFF
--- a/docs/ESP32.md
+++ b/docs/ESP32.md
@@ -55,17 +55,14 @@ To use it you need to have `#define USE_AUTOCONF`.
 
 ### CPU Temperature Sensor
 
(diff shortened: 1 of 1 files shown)


arendst:
SetOption127 still exists. You cannot just replace it if you happen to don't know it.......

alek67890:
I rewrited changes to include both commands. 
Also i added warning to not use that feature becouse of problems with him. I din't found anything about that problems in doc but there was many in issues for example #https://github.com/arendst/Tasmota/issues/12877#issuecomment-1155319190

PR #1127, opened 2022-12-13
Added ZPA GH305( SML)

MP1337:
New device added


MP1337:
Please add The Zpda GH305 (SML) device

PR #1139, opened 2023-01-01
Update DS18x20.md

aturlov:
Naming for multiple sensors in Rule's trigger uses defis - not the underscore _.


barbudor:
The correct description is that it depends on SetOption64
Please update the text to take SetOption64 into account
Thanks

PR #1144, opened 2023-01-13
Fixed unconverted links

heikokue:
Some links without regular markdown syntax are not recognized and converted by makedocs.
I put all of them into angle brackets.


heikokue:
fixed merge conflict

PR #1147, opened 2023-01-15
Update Sonoff-iFan03.md

bigwoof:
Added text to explain un-pairing of RF remote from ifan03 and a note on the possibility to control multiple ifan03 controllers with a single RF remote


bigwoof:
I found that the docs didn't note how to unpair the remote and also found that multiple controllers can be controlled by a single RF remote. 

Hopefully this will help others in the future.

PR #1148, opened 2023-01-15
Add revision 3 of sonoff 4ch device.

jowi24:
The information is the same for revision 2 and 3 of the device, so I suggest adding the information that revision 3 also exists.


barbudor:
Still templates is the best sites for that : https://templates.blakadder.com/sonoff_4CHR3.html

We should get rid of dupplicate information

PR #1149, opened 2023-01-15
Update EQ3-TRV.md

ariznaf:
I have met a problem trying to use TRV in an isolated network (with no access to internet). States of discovered valves were not being published each TRVPeriod as the device had no correct time. Configuring a local ntp server solved the problem.

I think that would be good to be comented in the documentation.


barbudor:
This is nothing related to EQ3-TRV
It is a basic requirement of Tasmota to have access to a time server and may be needsw to be clearly stated in the Getting-Started page

PR #1150, opened 2023-01-15
Update Bluetooth_ESP32.md

ariznaf:
TRVPeriod won't work if time is not correctly stablished in the device. NTP won't work at its default configuration if the device is connected to an isolated LAN. It would worth to mention under the TRVPeriod comments.


barbudor:
While I do agree that - if true- warning that properly set time is needed this is not a place for a lesson on how to get time.
There is surely a better place or even a dedicated page that could be referenced from here
Don't mix topics

PR #1158, opened 2023-01-23
Update DeepSleep.md

jsiddall:
Add a note that the calculation for the next wake time is based on the number of cycles since the epoch, not the start of the current day.

### DIFF
--- a/docs/DeepSleep.md
+++ b/docs/DeepSleep.md
@@ -5,6 +5,8 @@ DeepSleep support for up to 10 years (i.e., 86,400 seconds = 1 day) (e.g., if us
 !!! example
     With `DeepSleepTime 3600`, the device will wake up exactly every hour (e.g., 8:00am, 9:00am, ...). If you define `DeepSleepTime 86400` (i.e., 60\*60\*24), it will wake-up exactly at 0:00 local time. There is no option to shift the wakeup time; except changing the timezone. If you define `DeepSleepTime 600`, it will wake-up every 10 minutes (e.g., 8:00, 8:10, 8:20, ...).
 
+!!! note The next wake time will always be an even number of `DeepSleepTime` cycles since the epoch (Midnight 1970-01-01).  This may matter if the sleep time isn't an even number of minutes/hours (ex: 3660), such as when trying to wake at a specific time of day.
+
 !!! warning
   Please be aware that the minimum DeepSleep time is 10 seconds.
   


stefanbode:
Do you think we should change this. I assume incorporating the time zone is already in.

jsiddall:
My preference would be to change this.  I only came across it because things weren't working like I expected and I just did a quick division of the next wake epoch time (epoch time was my hint) by the `DeepSleepTime` and found that it was always a whole number.  I assume it was done this way for coding efficiency, but honestly I haven't even looked at the code.

The reason I would like to change this is that the current implementation makes it difficult to wake at a specific time each day, like an alarm.  I haven't found a trivial way to, for example, wake at 10:00.  If the next wake time were calculated from midnight of the current day, I think all the existing Tasmotas that are trying to wake every hour, minute etc. would keep working, but setting the `DeepSleepTime` to 36000 would wake at 10 AM local time as most people would expect.

However, an issue also arises in the 10 AM example.  Everything would work great at 10:00 and 20:00 but what happens when the next day's midnight comes around?  A simple calculation would result in 30:00, which would be 06:00 the next day, not 10:00.  So the code would need to check for the next sleep time crossing midnight and reset the calculation to be based on midnight tomorrow.  Not terribly difficult, just somewhat more difficult.  There could be other corner cases I haven't thought of as well (ex: the dreaded DST).  Time calculations never seem to be easy.

Regarding timezone, I also haven't checked that either, but I expect the calculation is based on local time not UTC.  I do recall seeing an issue about that along my searching.

I am not too good with C++ myself so not sure how much I could contribute to the effort but let me know your thoughts.

barbudor:
Deepsleeptime has always be about periodic wake-up, not wake up at a given time
The algorithm is based on that periodicity and would not work anymore if the wake up is expected at a given time with variable time between wake-up
May be the command should have been named DeepsleepPeriod but this is how it is intended

Deepsleeptime 36000 => 10:00, 20:00, 06:00, 16:00
is what it has been designed for

A wake up at specific given time would be a new feature request

stefanbode:
Anyhow there are some tricks you can use. First of all if you want to have a daily wake-up you need a deepsleeptime of 24x60x60=86400. now you wake-up at midnight every day. To shift this you can change the Timezone. Yes, this is a quite strange workaround. And you only can wake-up at full hours. But maybe this addresses your use case.

jsiddall:
Yeah, waking every hour is probably workable, just not ideal.  Having the ability to wake at times relative to the current day, not an arbitrary day more than 50 years ago, would be nice too.

stefanbode:
Please give it a try with the 86400 deepsleeptime (24h) and then change the Timezone. Let us know if it works

jsiddall:
As an off-board alternative to updating the way Tasmota works I did some statistical analysis of how close I could come to a desired wake time by choosing a specific `DeepSleepTime`.  I did an analysis of 1000 different times and found I could usually get within a few minutes of a desired time.  This makes intuitive sense since dividing 2 days into 1000 pieces would be about 3 minute intervals.  Since Tasmota clock is not that accurate a few minutes might be a reasonable.

As an example, if I wanted the Tasmota to come on at 04:11 UTC tomorrow morning, I could set the `DeepSleepTime` to `173383` and it would actually wake up at 04:13:17.

Maybe there is some formula that could solve for the best value but I am not a good enough mathematician to figure it out.  A relatively easy for loop is probably good enough for me.

jsiddall:
One more odd thing: the documentation for the `DeepSleepTime` command says the range of accepted values is:

> 0 = disable deep sleep mode (default)
> 11..86400 = set deep sleep mode time period in seconds

See: https://tasmota.github.io/docs/Commands/#deepsleeptime

But I know 10 is a valid time (and I also discovered it has special meaning, which maybe should be documented), and I have tested a value of 100,000 and that also seems to work fine.  According to the DeepSleep page the value could be up to 10 years, or 315360000 seconds.  Should the command documentation be updated to say from 10..315360000?

barbudor:
From the code, 10 is not a valid value. You can't enter `DeepSleepTime 10` as only value strictly greater than 10 and lstrictly lower than DEEPSLEEP_MAX are valid
You are maybe mistaking the impact of Teleperiod 10 in relation with DeepSleepTime
Teleperiod 10 (or 300) will insure that Tasmota will go back to deep-sleep as soon as possible right after the 1st Teleperiod message (which normally occurs ~ 12 sec after wake-up)

DEEPSLEEP_MAX is actually 316 224 000 (10 * 366 * 24 * 60 * 60) so yes, a PR on the doc to fix that would be appreciated

jsiddall:
Hmmm... sorry, I need to correct myself.  I should have said:

> But I know 10 is a valid **_TelePeriod_** (and I also discovered it has special meaning, which maybe should be documented)

I just tried and confirmed that trying to set `DeepSleepTime` to 10 simply results in it being disabled:

> 15:51:32.013 CMD: DeepSleepTime 10
> 15:51:32.018 RSL: RESULT = {"DeepSleepTime":0}

So I can updated the DeepSleepTime command, but then there is still a discrepancy in the Deep Sleep documentation, which says:

> Please be aware that the minimum DeepSleep time is 10 seconds.

https://tasmota.github.io/docs/DeepSleep/

Can we just delete that?  The deep sleep command documentation seems to have that covered.

barbudor:
In general we provide syntax and acceptable values in the Command.md page which is a reference
The specific page like DeepSleep.md should elaborate the usage of the feature in general and not dupplicating what is provided in the command description which should be assumed.

For example you could explain the specific usage of Teleperiod 10 / Teleperiod 300 with regarde to the DeepSleep feature in the DeepSleep page but not on the Teleperiod command.
On hte other hand it's on Command that you should find the allowed values : which starts at 11 for DeepSleep but which IMHO is way too low for a reliable operation.

barbudor:
By the way, adding more commits when hte PR as already been merged is probably useless
I'm not sure these commits can be merged
A new PR is probably needed

jsiddall:
#1160 is a new PR, the comments only reference this PR since this is where the confirmation of the correct range for `DeepSleepTime` exist.

stefanbode:
Hi, why do you use 10 seconds? This is not a use case. Normally 60 is a reasonable minimum. For sure you can do more than a day but this creates some of the shifts in time you mentioned. For sure 2days is also supported. Regarding the time shift: this is dependent of the temperature and voltage of the device. The deepsleep measure on every wake-up the error an compensate it. If the time is a day on constant temperature it will take about a week to be stable. Anyhow several minutes are normal on this long period. See documentation for examples of the time compensation. One idea easy to implement is a negative deepsleep to define the second of each day. So it is a daily wake-up, but at predefined time.

jsiddall:
To be clear, I wasn't saying I wanted 10 seconds, just that I want the documented range of values to match the code.  I thought the code allowed 10 seconds, but I was confusing the `DeepSleepTime` with the `TelePeriod`, which does allow 10 seconds.

11 seconds is effectively the same as 10 though, and that is allowed.  I also think that it is too low to be useful, except maybe for testing going in and out of sleep -- which it would do a lot!  When I was testing my DeepSleep code I used 20 seconds, and that worked great for testing.  But is it really worth changing the code to prevent people from using lower values?

jsiddall:
Update from my earlier test of 86400 `DeepSleepTime`.  Sleep started at just before 10 AM local time:

> 09:56:39.448 RSL: RESULT = {"DeepSleep":{"Time":"2023-01-26T00:00:00","Epoch":1674691200}

and the Tasmota did wake up around midnight local time and requested an IP address from the DHCP server, actually Jan 25 23:47:38 in GMT-5, but I suspect the approx. 12 minute error is just ESP clock drift.

However, it also woke up in the afternoon about 5 hours after it first went to sleep which was about 9 hours too early.  Odd.

I will watch it for another day and see what happens.

jsiddall:
Another update: not sure why by the Tasmota woke up a couple of times during the day at seemingly random intervals yesterday.  I just set this up temporarily with Dupont connectors but maybe they are a bit flaky.  Anyway, eventually it did way again just before midnight at Jan 26 23:50:37.  Looks like it is waking daily at midnight in local time.

barbudor:
Please let's keep the discussion abut Tasmota features on the Tasmota discussions and stick here about documentation
I agree that less 60 seconds DeepSleep in production doesn't makes sense although I often use 30 or 20 seconds when doing some test for having shorter cycles
I also imagine that any duration above a few days is probably useless too considering that there will be short wakes in the meantime has the ESP8266 is limited in real sleep duration

stefanbode:
Yes there are a a required wake-up every 60min. But beside a small power change and a 1 second up you should not see anything. At least there should be no wlan connect or any mqtt connect. The small wake-up is mandatory because of some limitations. See the documentation. If we allow 3 or more days we have this 1970 discussion most people do not understand. As mentioned a "deepsleeptime 86400 14:45" to wake-up daily at 2:45pm is something that may work easily. The variable is 32bit. If we restrict the deepsleep to 16bit 64.000 we can use the other 16bit for a shift of deepsleep without requesting an additional 16bytes in the permanent storage.

jsiddall:
Per @barbudor request  I created discussion #17813

@stefanbode, can you copy your last comment over there so we can continue discussion?

PR #1160, opened 2023-01-25
Update Commands.md

jsiddall:
Updated incorrect DeepSleepTime range

See discussion on #1158

### DIFF
--- a/docs/Commands.md
+++ b/docs/Commands.md
@@ -160,7 +160,7 @@ See also|[`SetOption1`](#setoption1) - Set button multipress mode<BR>[`SetOption
 Command|Parameters
 :---|:---
 Delay<a class="cmnd" id="delay"></a>|`2..3600` = set a delay between two backlog commands with 0.1 second increment. *Not recommended for precision timing!*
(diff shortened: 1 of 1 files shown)


stefanbode:
I do not support the merge. Create just confusion.

jsiddall:
The purpose of documenting the commands is so users know what they _can_ set a value to.  This is not the same as what they _should_ set a value to.  The values the user _should_ use can be documented on the DeepSleep page.

Right now the documentation does not match the code, which this PR resolves.

stefanbode:
Still disagree. Maybe we can use the values very high to do a mapping on the daily wake-up at a specific time. It is like an api definition. The definition ist what you also can expect tomorrow. Maybe today also some other stuff works but this can be changed tomorrow. Therefore no change for the max time of a day (24h).

jsiddall:
I am not sure I understand.  Are you saying you think the documentation _shouldn't_ match how Tasmota works, or that the way Tasmota works _should_ change?

As mentioned in the other DeepSleep thread, I can already see use cases for sleeping for 2 days and beyond, and if you were using a battery powered sensor monitoring very slowly changing data, I could see a use case for weekly or monthly telemetry.

Perhaps there is no need for a 10 year sleep, but I also don't see a need for arbitrarily preventing a user from setting the time longer if they want.

barbudor:
I think it's too old now and things may have change
Please resubmit a new PR if needed

PR #1162, opened 2023-01-29
Add SWB1 COM2 pinout

jrstarke:
I discovered that the COM2 through holes actually connected to the defined pins on the WT8266, with the exception of the 5V pin, which connects to the VCC pin through the voltage regulator.


barbudor:
Hi
I'm probably not well awake but I can't see in the photo anything labeled COM2
May be some arrows or indication would make it more easy?

jrstarke:
Did I assume badly that my photo with labels connected to the through holes indicating what each one did made it clear? 

Would it be clearer if instead of referring to it as COM2 (the tiny label so the top of the through holes) I just referred to its location (the open through holes on the far right edge of the board, indicated in the photo below)?

barbudor:
No, it's purely on me
I looked at it on my phone and for some reasons the right part wasn't visible
I said I was not fully awake

That's perfecr, my apologies

barbudor:
It might be interresting to add this WT8266-S1 module it to the Pinouts.md page

PR #1163, opened 2023-01-29
Mismatch Multicast Address in doc vs code

ld3300:
The Multicast address in the ArtNet doc does not match the address used in my_user_config.h (239.254.25.54 in code, 239.254.25.59 in doc)

Thank you!


barbudor:
Correct

PR #1173, opened 2023-02-10
Update MFRC522.md

momo90-2:
My ESP did not boot until I disconnected the RST connection. Somewhere this is also already mentioned as issue.

### DIFF
--- a/docs/MFRC522.md
+++ b/docs/MFRC522.md
@@ -27,7 +27,7 @@ This reader is ubiquitous in many Arduino starter and sensor kits. It uses SPI p
 |  RST    | GPIO0..5,15,16 |  RC522 Rst
 |  3V3    | 3V3            |
 
-If the ESP8266 doesn't boot, remove the RST connection.
+Warning : on esp8266, as the RST pin is pulling the signal high by default, gpio15 cannot be used for that signal. Please check esp8266 gpios specifications or the table in the section "Best Pins to Use – ESP8266" at https://randomnerdtutorials.com/esp8266-pinout-reference-gpios/
 
 ### Tasmota Settings 
 In the **_Configuration -> Configure Module_** page assign:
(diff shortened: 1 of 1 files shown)


barbudor:
Which GPIO was it connected to?
I don't know the characteristics of that RST pin but
GPIO 0 and 2 shouldn't be driven low at boot
GPIO 15 shouldn't be driven high

It would be better to clearly state which GPIO shouldn't be connected to RST

momo90-2:
I used GPIO15 for RST. I didn't try another one as it worked without problems after I disconnected the RST. I use it now without the RST connection.

Somewhere there is an issue that brought up the solution along with the explanation, but I can't find it anymore

blakadder:
GPIO15 is a restricted pin and should be avoided, doesn't have anything to do with the peripherals connection

momo90-2:
Well, then the docs should be changed, so that GPIO15 should not be used.
Now `|  RST    | GPIO0..5,15,16 |  RC522 Rst`
Should be `|  RST    | GPIO0..5,16 |  RC522 Rst`

And probably 
Now `|  SDA    | GPIO0..5,15,16 |  RC522 CS`
to `|  SDA    | GPIO0..5,16 |  RC522 CS`

blakadder:
GPIO15 I'd not the only restricted pin. Docs should be changed so they don't refer s specific  GPIO number

barbudor:
It's not so black and white
There are contraints at boot time on gpio0 2 and 15 that I have listed
That do not prevent using this lines but care should be taken to follow the rules
This is all explained in the docs about each gpio specificities

As for the MRFC, the information that might be added is
Warning : on esp8266, as the RST pin is pulling the signal high by default, gpio15 cannot be used for that signal. Please check esp8266 gpios spécificités in xxxx

Note that might also be true for the other signals if they have a pull up. 

Also don't make général assumptions that everyone will use the same ESP.

Please rewrite your submission
Thanks

momo90-2:
Changed it. I think the table from https://randomnerdtutorials.com/esp8266-pinout-reference-gpios/ is quite clear

barbudor:
Changed the link to tasmota documentation where all htis is already documented
No need to refer an external web site

PR #1176, opened 2023-02-11
Update AM2301.md

zaum:
AM2320 has 4 pins and will only operate in one wire state when the SCL pin is grounded.


blakadder:
thanks

PR #1182, opened 2023-02-17
Update DS18x20.md

dgoadby:
Add the comment - "Note: Experience has shown that the 4k7 is necessary even for a single DS18x20 sensor".  

If the 4k7 is added then this command is not required so maybe a better change would be to remove the reference to the command and to add a note making the 4k7 mandatory.


blakadder:
how much experience are talking about? dozen sensors? 50?

DeflateAwning:
Any updates on this? Seems reasonable to merge with this warning imo; otherwise, is there a better wording you'd like to see?

PR #1223, opened 2023-04-28
Add description of configuring a switch for autodiscovery.

markfickett:
This method works for me quickly/easily. The default instructions only make something show up in the UI if I use the relay, which is not recommended. And there are much more complicated workarounds described, for example at https://blakadder.com/pir-in-tasmota-integration/ , and a number of threads one can find with people trying to figure out how to do this. So I think it would really help to have this method described in the main PIR docs.


blakadder:
having a motion sensor in HA as a switch is wrong, it's a binary sensor

markfickett:
I can't argue with "switch" being the wrong terminology. But it did show up as a binary sensor in HA (with entity name "binary_sensor.switch1"), different from the on/off control for the LED attached to the ESP8266 which instead shows up as a switch.

image

PR #1228, opened 2023-05-06
Fixed syntax of unordered lists in TasmoAdmin chapter

mundanevision20:
I'm pretty new to Tasmota. Reading the docs I noticed that the syntax of the unordered lists does not seem to match the guidelines given in the [Markdown guide](https://www.markdownguide.org/basic-syntax/#unordered-lists). Due to the missing spaces the indented items were not rendered correctly.


mundanevision20:
I do noticed that [markdownlint](https://github.com/DavidAnson/markdownlint) advises to use two spaces but mkdocs requires four.

PR #1239, opened 2023-06-12
Update ADC.md

Darthmonkey:
Fixed some spelling and grammatical errors, nothing technical.


blakadder:
thanks

sfromis:
The technical change you did do is not quite correct, with Tasmota, the ESP32 *chip* has eight ADC inputs, not just four.

PR #1243, opened 2023-06-22
Update Thermostat.md

KDMcMullan:
Grammar corrections to clarify "bypass". Other simplifications.
Corrected default timeresetset value to match the suggested user_config_override.h.


KDMcMullan:
Just for clarity, the sentence _"If the application requires to follow the command of the input once active (see bypass setup explained in previous section)."_ was the one I was targeting, but I did s couple of others, too. Have I done too much for a single change?
Grasping "bypass" and the polarity of **cmnd/Tasmota_Name/INPUTSWITCHUSE** were the things I struggled with most when exploring thermostat.

blakadder:
care to check the accuracy of these additions @arijav

mendel5:
I think this is a useful change and would appreciate it if somebody could merge this pull request.
However, we need to consider the merge conflict that is caused by this change: https://github.com/tasmota/docs/pull/1295 `Replace "f.e." with "e.g.,"`
So we just need to make sure to keep the `e.g.`.

mendel5:
Thank you!

PR #1244, opened 2023-06-22
Update Gitpod.md

KDMcMullan:
Changed ' "/tasmota" root directory' to avoid time spent looking in "/tasmota".

### DIFF
--- a/docs/Gitpod.md
+++ b/docs/Gitpod.md
@@ -45,7 +45,7 @@ To modify the stock configuration:
 
 ### Prepare the IDE for Compilation
 
-1. Open the `platformio.ini` file located in the `/tasmota` root directory (scroll to the bottom of the file Explorer (1) pane). In this file, removing a leading semicolon `;` enables a statement.
+1. Open the `platformio.ini` file located in the tasmota root directory (scroll to the bottom of the file Explorer (1) pane). In this file, removing a leading semicolon `;` enables a statement.
 2. In the Editor (2) pane:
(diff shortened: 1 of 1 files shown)


blakadder:
that makes no sense

KDMcMullan:
Quite correct. On reflection, it would make more sense to change it to "/TASMOTA root directory". 

The guide is an excellent piece for someone who's not used GitPod before nor compiled TASMOTA before. As such a person, I followed the guide. Unfortunately the /TASMOTA directory contains a folder called tasmota, so I spent some minutes looking for platformio.ini in /TASMOTA/tasmota, when I should have been looking in /TASMOTA. In my fluster, I missed the instruction to scroll to the bottom.

There's an inconsistency, though. Elsewhere in the document, we are instructed to create user_config_override.h in the /tasmota folder. In this case, I'd suggest it changes to /TASMOTA/tasmota .

If you wish to reject the change, I'll gladly make a more appropriate suggestion. Do you prefer:

_Select the /tasmota folder in the Explorer (1) pane. Create a new file called user_config_override.h_
(which could be perceived wrongly as it's actually /TASMOTA)

or

_Open the platformio.ini file located in the /tasmota root directory_ 
(which could be perceived wrongly as it's actually /TASMOTA/tasmota)
 
Thanks for looking.

barbudor:
May be just 
  located at the bottom of the file Explorer pane.

And add a picture of that:
image

PR #1263, opened 2023-08-02

bababue:
Hey everyone,
this is my first pull request, but it also is just a very small change.
For some reason the SMI script for the KAIFA MB310H4BDE did not have a # at the end which broke its functionality. All other scripts have a # and it is requiered to make the script work.

Thanks for considering this PR and have a nice day


blakadder:
thanks

PR #1276, opened 2023-09-08
Update IAQ.md

chrfriese123:
iaQ-Core no longer produced, change to ENS161 from same manufacturer


blakadder:
irrelevant to the driver

PR #1279, opened 2023-09-15
Added requirements for WebQuery

pmknowles:
Added the requirement to compile custom bin using #define USE_WEBSEND_RESPONSE if a response is required from a WebQuery.


barbudor:
Thanks @pmknowles

PR #1283, opened 2023-09-24
Add sonoff s31 docs for prom units mismatch

subtlepseudonym:
## Summary
As mentioned in [this issue](https://github.com/arendst/Tasmota/issues/19538), on the Sonoff S31, the prometheus metrics `energy_power_kilowatts_daily` and `energy_power_kilowatts_total` report values whose units are kilowatt _hours_ rather than kilowatts, as implied by the naming convention.

This change adds an update to the Sonoff S31 device documentation to point this out.

### Notes
Ideally, we'd update the metric names, but those metrics are shared across quite a few devices so implementing the name change would be quite a bit of effort for a pedantic fix. In the mean time, folks running into this issue can work around it with [metric relabeling](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#metric_relabel_configs).


Jason2866:
Thx!

PR #1286, opened 2023-09-29
Update Getting-Started.md

helgek:
Hi,

I added one paragraph as a proposal to make visitors aware of the existence of an outstanding tool called PCBite which allows flashing without soldering for those microcontrollers where there's no other option otherwise to flash.

Please consider my post in this [forum](https://github.com/syssi/esphome-yeelight-ceiling-light/discussions/72) for more details.

Best,
Helge

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -34,6 +34,8 @@ To solder you'll of course need a soldering iron, soldering tin and some flux. I
 
 If you're intimidated by soldering there are 3D printed jigs available for different modules and devices. At worst, you could get away with holding the headers tightly with jumper wires in pin holes during flashing but it is not a foolproof process and flashing might fail.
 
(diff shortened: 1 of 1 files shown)


blakadder:
No

Jason2866:
No product placement

helgek:
@blakadder @Jason2866 When there's only one solution on the market to provide the opportunity to many people not gifted with soldering skills to eventually flash their devices, how would you then empower them with the knowledge of this solution? I have no relationship with this company other than that I purchased their solution and successfully used it for four yeelight lamps which partly didn't work reliably (even in local network mode) and eventually through this solution I was able to flash them. Certainly the world will continue to exist also without people being able to flash some of their smarthome devices, nevermind if you don't want to add the reference.

blakadder:
Definitely not in the Getting started documentation of an open source firmware project

Jason2866:
There are nice alternatives. Open Source too. An example https://www.superhouse.tv/36-3d-printed-pcb-workstation-using-acupuncture-needles/

helgek:
@Jason2866 That comes very close indeed, thanks for sharing.

PR #1291, opened 2023-10-07
Add note for ADC CT Power users

cendern:
Users looking to calibrate their clamp ammeter devices (CT Power module hooked to ADC pin) can't use the commands on the page titled "Power-Monitoring-Calibration" because the PowerSet/VoltageSet/etc. commands don't apply to ADC-based measurements. Instead, those users need to calibrate using the `adcparam` command.

This PR adds a note pointing them in the right direction.

Hopefully this helps people to find what they want and avoid filing issues like these:
https://github.com/arendst/Tasmota/issues/8011
https://github.com/arendst/Tasmota/issues/7672


blakadder:
thanks

PR #1294, opened 2023-10-29
Fix script for EMH eHZ Generation K, part 2

mendel5:
Fix script for EMH eHZ Generation K, part 2


mendel5:
I thought about the naming of the variables again and came to the conclusion that the current names are better. Therefore I closed this pull request.

PR #1295, opened 2023-10-30
Replace "f.e." with "e.g.,"

DeflateAwning:
"e.g." is the correct abbreviation of "for example" in English


mendel5:
I think this is a useful change and would appreciate it if somebody could merge this pull request.

stefanbode (approved):

barbudor:
It doesn't build
There seems to be wrong link to pictures
Probably not your fault but it would be great to fix it before the merge

DeflateAwning:
Thanks for checking it over and merging. I just branched off of master, so that's likely where the issues came from

mendel5:
Thank you!

PR #1297, opened 2023-11-03
Replace "KWh" with "kWh"

mendel5:
Replace "KWh" with "kWh" because the correct abbreviation of a kilowatt-hour is "kWh" with a lowercase "k", see https://en.wikipedia.org/wiki/Kilowatt-hour


mendel5:
Hi @blakadder , I would appreciate it if you could merge my pull request. If you have any questions, please let me know.

mendel5:
Thank you, @stefanbode

PR #1307, opened 2023-12-01
MAX7219 dot matrix docu extended

sternegugger:
The format is now more readable. Some corrections and better explanations have been made.


blakadder:
the compile information has to be included at the top of the page as the rest of the sensors see for example https://tasmota.github.io/docs/AHT2x/

PR #1308, opened 2023-12-04
Update Timers.md

stheine:
I struggled with the Timer configuration UI and the documentation, so I tried to make this a little more understandable.


blakadder:
it's less understandable to me

stefanbode (approved):

stefanbode:
But it is correct. Depending on the mode there is a different behavior.

PR #1309, opened 2023-12-06
Added M1P_V1.0 module (ShellyPlugUS) to Pinouts.md

jdwhite:
The M1P_V1.0 is an ESP8266 module used by the ShellyPlugUS. In addition to pinout information, the accompanying images show the correct pins for TX and RX as the ShellyPlusUS board is mislabelled.


barbudor:
Hi, thanks for the update
Just wanted to check a point regarding 'pins mislabelled'
Do we agree that your picture shows the RX and TX from the module/ESP point of view?
i.e. The RX pin of the module is the serial receive to be connected to the TX pin of the USB adapter and TX to RX?

jdwhite:
Thank you for your skepticism. I don't blame you one bit because this can be a point of confusion and I would also question this claim.

> Do we agree that your picture shows the RX and TX from the module/ESP point of view? i.e. The RX pin of the module is the serial receive to be connected to the TX pin of the USB adapter and TX to RX?

Yes, this is the case and I was also surprised by this. I've been fiddling with RS232 for over 35 years and flashed many other ESP modules and it's always been module RX <-> USB TX, module TX <-> USB RX for DTE equipment (and I claim it still is and that these pads are mislabelled). After trying unsuccessfully to flash using typical RX<->TX, TX<->RX I swapped TX and RX and it worked. Go figure.

In the spirit of "pics or it didn't happen", here's a pic of it wired up. I've color coded everything the same from the Shelly to the USB/TTL pins. Hopefully you can see the TX/RX markings on the Shelly and the corresponding USB/TTL pins. This will confirm the RX<->RX and TX<->TX wiring. The screenshot in the background is the Tasmota log output from the module. I can also provide a video of me flashing Tasmota in this config, which would confirm two-way communication, if needed.

I realize that my TX/RX connections aren't connected to the module pads as VCC/GND/GPIO0 are, but instead to the marked TX/RX pads that are a few mm away, but they are electrically identical per DVM continuity test. Guessing the manufacturer put those pads there for convenience.

I'm thinking I should also add the close-up wiring diagram photo to the pull request as I'm sure there will be others who question my claim. Was also considering pointing out the "short GPIO0 to ground for 5-10 seconds while energizing to put into programming mode", but that seems pretty standard and as this is a page about pinouts so could argue it doesn't belong here.
Thoughts?

M1P_V1 0 ShellyPlugUS Wiring

M1P_V1 0 ShellyPlugUS Log

barbudor:
À simple 'yes' would have been enough 🤣🤣🤣

Personnaly I would consider that the last photo of the PR is enough but all good for me

Thanks

PR #1311, opened 2023-12-08
Update Getting-Started.md

jdwhite:
Include warnings regarding mislabeled TX/RX solder pads on ShellyPlugUS circuitboard as referenced in https://github.com/tasmota/docs/pull/1309

### DIFF
--- a/docs/Getting-Started.md
+++ b/docs/Getting-Started.md
@@ -111,6 +111,8 @@ When you have identified pins on your device, connect wires according to the tab
 **Note that TX from your adapter goes to RX on the ESP device and RX from adapter goes to TX on the device!**
 ![Image courtesy of https://www.domo-blog.fr/](https://user-images.githubusercontent.com/5904370/57880182-69bf2f80-781e-11e9-8953-88599cb89155.png)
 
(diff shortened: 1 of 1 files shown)


blakadder:
I don't see how the pins are mislabelled in the case of shelly plug if there are no labels on the PCB

jdwhite:
I’m referring to the pics in the discussion on 1309, not the PR. In the PR I blocked out the mislabeled pads on the shelly PCB. Apologies if that caused any confusion. -Jason-- Jason White“The single biggest problem in communication is the illusion that it has taken place.” - George Bernard ShawOn Dec 12, 2023, at 6:26 AM, Jason White ***@***.***> wrote:﻿
Please look at the pictures in 1309. You will see TX and RX on the PCB. The pictures show how I had to wire up the Shelly to the serial adapter. -Jason-- Jason White“The single biggest problem in communication is the illusion that it has taken place.” - George Bernard ShawOn Dec 12, 2023, at 2:53 AM, blakadder ***@***.***> wrote:﻿

I don't see how the pins are mislabelled in the case of shelly plug if there are no labels on the PCB

PR #1312, opened 2023-12-11
Add documentation page for MAX31855 thermocouple amplifier and link from various other locations.

markfickett:
I also moved the `USE_MQTT_TLS` comment inline to the table, since the lone asterisk was making formatting off in my editor and it's only referenced in one place.


docs/MAX31855.md
```diff
@@ -0,0 +1,66 @@
+# MAX31855 / MAX6675 Thermocouple Amplifier
+
+??? failure "This feature is not included in precompiled binaries"
```
markfickett, line 3: This isn't displaying right when I "view file", I'm not sure if I have an error in the Markdown, or it's just a special feature that doesn't show in the PR view.

docs/BUILDS.md
```diff
@@ -15,7 +15,7 @@ Note: `minimal` variant is not listed as it shouldn't be used outside of the [up
 | USE_DOMOTICZ              | - | x / x | x | x | x | - |
 | USE_HOME_ASSISTANT        | - | - / - | - | - | - | - |
 | USE_TASMOTA_DISCOVERY     | x | x / x | x | x | x | x |
-| USE_MQTT_TLS*             | - | - / x | - | - | - | - |
+| USE_MQTT_TLS              | - | - / x | - | - | - | - | USE_MQTT_TLS is enabled by default in every ESP32 variants
```
barbudor, line 18: Please remove any changes from BUILDS.md

BUILDS.md and I2CDEVICES.md are actually automatically copied from Tasmota repo and should not be edited in the docs repo
To update them, please fill a PR on arendst/Tasmota

Thanks

docs/MAX31855.md
barbudor: markdown renderer included in editors such as VSCode are not exactly identical to the markdown renderer used in MkDocs so they may not be appear as expected in your editor while being perfectly correct.

This looks like fine to me

barbudor (changes requested):

docs/BUILDS.md
markfickett: Got it, reverted `BUILDS` and `I2CDEVICES`. Since I added links from a couple other locations I probably will just leave those as-is, but good to know where they come from.

PR #1317, opened 2023-12-17
Replace aambiguous = sign by : in rules

barbudor:
Hi @blakadder 
Following this mis-interpretation in https://github.com/arendst/Tasmota/issues/20251 I suggest to use `:` instead of `=è to separate the token from the explanation.
Done htat for Rules commands here
Can slowly start to apply to other blocks later on

Agreed ?


blakadder:
So you're changing one character with another so people can now use : as part of their misinterpreted syntax?

barbudor:
Ples adding a couple of extended exemples

I think that in the contect of Mem/Var, `=` is more ambiguous than `:` as they may be used with `=` in programming language.
And in general I feel `:` stands better to be followed by an explanation than `=`

blakadder:
= was there from inception of the wiki. Changing it to insignificantly reduce ambiguity is not a preferred outcome, Either remove ambiguity completely or leave it be. You will always have individuals skipping the intro on how to use commands

PR #1318, opened 2023-12-18
Update Components.md

andreasloe:
see https://github.com/arendst/Tasmota/discussions/20263


s-hadinger:
I understant it's an old PR, but I feel the wording is hard to understand. I will try to reformulate

PR #1323, opened 2023-12-24
Update Rules.md

pille:
add cookbook rule using power LED to show power consumption as brightness


sfromis:
That cookbook example does not quite work as expected, as `%var1%` is being expanded before the rule starts, meaning that it will use the prior value.

PR #1336, opened 2024-01-02
Update MQTT.md: format "mosquitto_pub" as command

bam80:
Otherwise, it's not obvious _mosquitto_pub_ is a command.


barbudor:
Thanks

I even added "command-line tool"

PR #1399, opened 2024-07-13
Update Displays.md

fb-pilot:
update links to https://tasmota.github.io/docs/Universal-Display-Driver/

### DIFF
--- a/docs/Displays.md
+++ b/docs/Displays.md
@@ -430,11 +430,11 @@ The EPD fonts use about 9k space, which can be selected at compile time using \#
 
 ## Universal Display Driver
 
-Documentation for Universal Display Driver was moved to a [specific page](Universal-Display-Driver/#universal-display-driver-udisplay)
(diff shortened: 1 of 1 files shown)


sfromis:
Why do you want to make those two links absolute, when the previous relative links (this format is used in all other docs links) are how it is supposed to work? Anyway, not long ago those two links were broken, I fixed them to the standard format, as used for other links.

IMO, this change should be reverted, without good reasons to do it.

arendst:
OK. I'll revert although the links then don't work for me.

arendst:
The reverted link ``https://tasmota.github.io/docs/Displays/Universal-Display-Driver/.md#universal-display-driver-udisplay`` doesn't work for me.

sfromis:
Made another fix, the issue was having a slash after the page name, before the `#`. Tested to work, even if what's shown when hovering the links in a browser is what does not work in the md files.

arendst:
Yep. Verified. Works fine now.

PR #1406, opened 2024-08-23

hagenbuch:
Dots to backticks, sorry I have to idea where the dots came from in my previous pull request.

BTW, I suggest putting all the metering scripts in a separate folder (one file per manufacturer) so the docs don't get too cluttered.

Thanks for your great work!


sfromis:
Oops, independently of your PR here, I noticed that formatting issue on the docs page, and fixed it a few minutes ago. Hence no need for the PR now.

hagenbuch:
Fantastic, many thanks for your work!

PR #1433, opened 2024-12-05
Fixed Language on Securing-your-IoT-from-hacking.md

mrdvt92:
I fixed some of the language in this document but, it needs much more attention.  Tasmota security will become more public now that the U.S. FCC has an IoT Cybersecurity Labeling Program.


stefanbode:
👍

PR #1442, opened 2024-12-29
fix: eBZ DD3 IO pin

Helge37:
Changed IO pin from 13 to 3 to be in line with the other examples.


arendst:
Why this change from GPIO13 to GPIO3???

GPIO1 and 3 are used for serial console. GPIO13 is an addition serial input having additional UART hardware support.

PR #1446, opened 2025-01-16
add "hidden" features to TOC

mi-hol:
without an entry in Table of Content (TOC) important features are easily overlocked. Align with other content


mi-hol:
for a reference manual the TOC for scripting language is extremely brief.
image
After above PR is accepted, I'd be happy to extend the TOC further

PR #1460, opened 2025-02-02
fix #1449 , but has dependency on mkdocs 1.6

### LINKED ISSUE #1449: Documentation "Search" for script section (ie. >D) does NOT show existing content
Example:

Searching for script section gives "no matching documents" despite the search string being shown in Table of Content on the right.
Maybe a configuration issue related to special characters like ">" is causing this issue?

Image

mi-hol:
the script command sections consist of only 2 characters (i.e. >D), hence with default search setting they will not be reported as hits. See https://www.mkdocs.org/user-guide/configuration/#min_search_length

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -9,9 +9,10 @@ docs_dir: docs
 
 plugins:
   - search:
+      min_search_length: 2
       lang:
         - de
-        - en
+        - en       
   - alternate-link
   - minify:
       minify_html: true


mi-hol:
it seems mkdocs needs to be upgraded to 1.6 for this setting to work.

https://github.com/tasmota/docs/blob/master/requirements.txt currently specifies 1.5.3

Who could test the update as a prerequiiste to my PR?

mi-hol:
upgrading is described in https://www.mkdocs.org/about/release-notes/

Staars:
I did some quick local tests and not encounter problems with the newest version.
Let's see ...

Staars:
It must be another problem as mkdocs 1.6.1 was already used anyway according to the GH action log:
```
Collecting mkdocs>=1.5.3,~=1.5 (from -r requirements.txt (line 2))
  Downloading mkdocs-1.6.1-py3-none-any.whl.metadata (6.0 kB)
```

Did you check successfully with a local installation?

mi-hol:
> Did you check successfully with a local installation?

No, I'm used to update docs in projects that use GH Actions to validate all changes.
The result is a temporary site that can be used to verify changes manually too.
From my perspective this approach is much more reliable because of all sort of issues in local installations due to version differences & conflicts.

mi-hol:
could the issue be caused by material-search?
This search has different options & severe issues as desribed in https://squidfunk.github.io/mkdocs-material/blog/2024/08/19/how-were-transforming-material-for-mkdocs/#search-and-discovery

Staars:
Yes, could be.
If you would have checked your PR with a local build, you would have seen an error like that:
```
mkdocs build --strict
WARNING -  Config value 'plugins': Plugin 'material/search' option 'min_search_length': Unsupported option

Aborted with 1 configuration warnings in 'strict' mode!
```
You see that the search plugin from mkdocs-material is referred here, which does not seem to have the option `min_search_length `.
Hence my suggestion to test such things locally first and not pushing untested things upstream.

Maybe someone with more knowledge about MkDocs can enlighten us.

mi-hol:
> Hence my suggestion to test such things locally first and not pushing untested things upstream.

well my goal is NOT to understand all internals of mkdocs, it just appeared to be an easy fix to an annoying issue in doc search.
As we know now its unfortunately not that simple :(

mi-hol:
@Staars @arendst maybe switching to [theme "Read the Docs" ](https://about.readthedocs.com/?ref=readthedocs.org) would resolve the [severe issues mention by theme "material" developers]( https://squidfunk.github.io/mkdocs-material/blog/2024/08/19/how-were-transforming-material-for-mkdocs/#search-and-discovery)
and would lower the hurdle to documentation updates significantly!
At the same time quality of doc updates would increase via Automatic deploy previews
image

PR #1474, opened 2025-04-15
Update UFS.md

evidalr:
Added **Beware, only filnems up to 31 characters long are allowed.** to the description of the Upload button.


evidalr:
Sorry, I wrote "filnems" instead of "filenames".

PR #1482, opened 2025-04-29
Update Smart-Meter-Interface.md - clarified option so3

gone-for-coding:
Clarified special option so3 to make it more easy for the user to identify the number of arguments and the defaults.


gone-for-coding:
Discussion on how this option works can be found here https://github.com/arendst/Tasmota/discussions/23342.

PR #1487, opened 2025-05-14
Remove unnecessary library setting

Noschvie:


s-hadinger:
I'm wondering, maybe you need both `lib_audio` and `libesp32_audio`

Noschvie:
With these settings, it compiles without errors. But I can't test it until this evening.

```
lib_extra_dirs              = ${common.lib_extra_dirs}
                              lib/libesp32
                              lib/libesp32_audio
```

Noschvie:
https://github.com/arendst/Tasmota/discussions/16226#discussioncomment-4750642

s-hadinger:
Oh you are right. There was a ghost `libesp32_audio` directory locally on my machine, but it's empty indeed.

Noschvie:
Fixed to lib/libesp32

Staars:
The current documentation is still correct.
lib/lib_audio is needed and the only specific lib folder for audio. But it may be part of the default lib folder list, which is not guaranteed forever. Thus the user must ensure, that it is included.
lib/libesp32 is of course needed for audio on the ESP32 too, but it is always mandatory for every ESP32 build. There is nothing audio related in it and it is pretty much guaranteed, that it is part of the default lib list for ever.

Staars:
There is really no guesswork or surprising stuff happening.

The current included library folder are defined here:
https://github.com/arendst/Tasmota/blob/28f4a07fd611d9c04ad2b5ef684e927f1ae0476d/platformio.ini#L48-L55

Unless someone completely overrides the entry `lib_extra_dirs` in a projects build section, then these libs are always compiled. At the moment the audio libs are included and thus compiled with every ESP32-xx-build.
I would even say, this should NOT be the case for such a niche application, because every build will be slowed down by the audio lib compilation. But this is another story.

The title of this PR is not correct and you could rename it to something like: `remove unnecessary library setting`
But keep in mind, that the current default is subject to change and it might be necessary to manually include the audio libraries to `lib_extra_dirs` for an audio related project again.

Noschvie:
Now I have deleted the line.

Noschvie:
If desired, I can remove the lib/lib_audio line from the code and adjust the documentation accordingly. If I understand correctly, there's no difference between the ESP8266 and ESP32 in this regard, right?

PR #1488, opened 2025-05-15
Remove unnecessary library setting

Noschvie:
Like PR #1487 for ESP32


Staars:
Not really.
This whole page is a remnant from the old driver version before Arduino 3.x.
It was left for reference (and should be unlinked)  mainly for the ESP 8266, which has very limited audio support. I have zero experience with it.

PR #1489, opened 2025-05-17
Updated esptool docs

ngehrsitz:
Noticed that the commands here are still using the deprecated format with `_`: https://docs.espressif.com/projects/esptool/en/latest/esp32/migration-guide.html#command-and-option-renaming
Also backups should read `ALL` instead of only 1M of the flash


sfromis:
As esptool v5 is not even released yet, I'd say that it is too early to change underscore to hyphen, as hyphen is not even accepted by the latest release 4.8.1. They say that v5 will be released this month, but not yet.

As using underscore is going to be tolerated even after deprecation, and many users will not quickly get to the new v5 right away (like what's included in various distros and toolchains), I'd even say that we *should* not confuse users by changing to hyphen until some time after the release of v5.

PR #1496, opened 2025-06-27
Fix some typos in Berry.md

dwrobel:


s-hadinger:
Thanks, yes we assume en_US.

I'm ok with most changes except:
- "filesystem" instead of "Filesystem" - I don't understand why you capitalized it
- "Zigbee" instead of "ZigBee" - the camel case here is ugly and quite uncommon

Can you please change accordingly?

dwrobel:
>I prefer to keep it lowercase in text.

PR has been updated accordingly.

s-hadinger:
Thanks

PR #1498, opened 2025-07-23
Update AM2301.md - fix GPIO default to stop null reads

dodge107:
Thank for the great work in this project!

Edited GPIO selection. If using GPIO2 for the sensor, and soft rebooting the unit all you get is null readings. Saw the issue on a number posts on line after I had the same issues follwoing the default details. Easy fix is to use other GPIO pins.

### DIFF
--- a/docs/AM2301.md
+++ b/docs/AM2301.md
@@ -60,7 +60,7 @@ Like the [Wemos DHT11 shield specs](https://docs.wemos.cc/en/latest/d1_mini_shil
 ### Tasmota Settings
 In the **_Configuration -> Configure Module_** page assign:
 
-1. GPIO2 to `AM2301 (2)`
+1. GPIOx to `AM2301 (2)` - Do not use GPIO2 as reading will not be available after a soft reboot.
 
 ## Sensors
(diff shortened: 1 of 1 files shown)


arendst:
On some ESP8266 GPIO02 is wired to a blue led which might interfere with the AM2301 interface. Anyway, thx for edit.

18923:
Years ago I resolved this "NULL" issue on an ESP8266, ESP01s by connecting the DHT22 Vcc to RX pin for power,
and using "Rule1 and Rule2" to cycle power the DHT22 sensor on boot/reboot.

DHT22 Vcc power connected to RX pin (GPIO3) and Sense pin to GPIO2.

Tasmota Configuration.

Configure module GPIO2 as AM2301 and GPIO3 as Relay1
Console settings
Set "seriallog" OFF _**(may not be necessary)**_
Rule1 on system#boot do backlog power1 0; ruletimer1 20 endon  **_#Turn power1 OFF 2 seconds after boot_**
Rule2 on system#boot do backlog power1 1; ruletimer1 60 endon  **_#Turn power1 ON 6 seconds after boot_**
Rule1 ON
Rule2 ON     **_#to enable rules......_**

# 82Xats

PR #1503, opened 2025-08-22
Update Smart-Meter-Interface.md with official FNN docs for OBIS codes

markusdd:
With more and more meters entering the market often times finding the exact complete script here is not possible, but they all seem similar. After reading the manual of my new meter I realized that there is a common document many meter manufacturers now follow for their datapoint codes. I think this can be helpful to others.


mi-hol:
@gemu2015 I find the addition valueable. Would you mind to review?

PR #1528, opened 2025-12-17
docs: Document that bus 2 is actually broken for most I²C peripherals

ntninja:
Sucks, but true.


arendst:
Manpower. Commitment. Tenacity.

arendst:
Yes.

If you're clever you'll make sure the Tasmota repository version is automatically copied/synced with the docs version. Otherwise it will run out-of-sync very soon ;-)

ntninja:
I actually meant just deleting the Tasmota repo version or having it state “See \<link to docs/I2CDEVICES.md>”…

PR #1553, opened 2026-04-20
Update TLS.md

hkspks:
Added instructions to reset the TLS fingerprint - a common issue, if the fingerprint changes due to server reconfiguration.

### DIFF
--- a/docs/TLS.md
+++ b/docs/TLS.md
@@ -34,6 +34,13 @@ or
 #define MQTT_FINGERPRINT2      0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00,0x00
 ```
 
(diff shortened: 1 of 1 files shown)


s-hadinger:
I'm converned. If you zero both fingerprints, then you make your device very vulnerable. Your legit server will be recognized as first fingerprint. When a rogue server shows up, it will take the second slots, hence defeating the purpose.

You should keep the original fingerprint and zero only one of them at the same time.

hkspks:
Thanks @s-hadinger, let's add this to the doc just to remove one fingerprint. For my interest: How are the fingerprints set at a fresh installation? I wanted to trigger a "factory reset". Furthermore a speaking error message would be great on the console, if the fingerprint of the server has changed. I got the following message and was kind of lucky to find the source of the problem:

`11:50:18.211 MQT: TLS connection error: 1
11:50:18.213 MQT: Connect failed to <server-ip>:8883, rc -2. Retry in 20 sec`

PR #1554, opened 2026-04-24
Add detail to INA226 commands

pkkrusty:
Adds clarification that INA226 channel directly maps to which I2C address it's on.


arendst:
Thx

PR #1555, opened 2026-05-15
Update TLS.md certificate requirement section for ECDSA

ntninja:
ECDSA was enabled even for ESP8266 since this was last updated.


ntninja:
@arendst: Can you verify this matches the current details in Tasmota?

PR #1564, opened 2026-08-09
ADC.md: state that AdcParam is now superseed by adcGpio

evidalr:
AdcParam is now superseed by AdcGpio.  The doc should be updated using AdcGpio instead of AdcParam.  I plan to do it, once I can be sure enough about the current meaning of the parameters under all the configurations of the Analog pin.


evidalr:
I am testing the (old AdcParam and the new) AdcGpio on an ESP8266 Wemos D1 board.  So far I have tried only with two template configurations of the (GPPIO17) Analog pin:  "ADC Range" and "ADC Input".  

For ADC Range, the four parameters of AdcGpio are actually  as explained in the ADC.md doc: 

 param1  = input range low value adcLow (default = 0)
 param2  = input range high value adcHigh (default = 1023)
 param3  = output range low value rangeLow (default = 0)
 param4 = output range high value rangeHigh (default = 100)

But for ADC Input, as far as I can tell, the meaning is diferent; namely:

 param1  = input lowest value adcLow (default = 0)
 param2  = input highest value adcHigh (default = 1023)
 param3  = pecentage of param2-param1 (default 5)
 param4 = must be 0 -- otherwise, analog input becomes disabled.

An (up or down) input change higher or equal to param3 (in %) triggers 
analog#A0div10, with %value% equal to the ADC input as a percentage with respect to adcHigh - adcLow.

I find this interpretation for ADC Input very convenient, since it grants a lot of flexibility for the Analog#A0div10 trigger.   However I would like to confirm that this is an intended feature and not a bug, with respect to what is said in de docs -- because it differs considerably from what is now documented in several places; for instance, in Rules  (and also in Analog pin docs) we read:

Trigger Analog#A0div10 -- Occurs when the A0 input changes by more than 1%.  It provides a value between 0 and 100

But it is NOT 1%, but param3 % instead; i.e., 5% by default.

Any comments welcome!

arendst:
I'll need time to review...

evidalr:
Take your time... Thanks.

